PR #1, opened 2024-01-29
zkcq + cleanup

arifiorino:
1. Strengthening the security of cq by having the prover send the elements `B_0_gamma, f_gamma, A_0` encoded as group elements rather than their plain values
2. Added default implementations to basic_block.rs to clean up the code. For example basic_block/add.rs doesn't need to implement the setup function as it has no setup.
3. Changed proof format to no longer include field elements as it is unsafe and no longer required
4. Added .gitignore
5. Ran rustfmt

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+target
+Cargo.lock
+.DS_Store
+*.swp
--- a/src/basic_block.rs
+++ b/src/basic_block.rs
@@ -1,3 +1,4 @@
+#![allow(unused_variables)]
 use ark_poly::univariate::DensePolynomial;
 use ark_poly::{GeneralEvaluationDomain, EvaluationDomain};
 use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine};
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -0,0 +1,4 @@
+target
```
ddkang, line 1: if you don't have a project wide rust format, I'd recommend adding one

src/basic_block.rs
```diff
@@ -38,23 +39,31 @@ impl DataEnc{
 pub trait BasicBlock{
   fn run(model: &Vec<Fr>,
```
ddkang: I generally prefer

function(
  arg1
  arg2
) -> return {
  stuff
}

for visual separation but up to you

src/basic_block.rs
```diff
@@ -38,23 +39,31 @@ impl DataEnc{
...
          inputs: &Vec<Vec<Fr>>) ->
-         Vec<Fr>;
+         Vec<Fr>{
+    Vec::new()
+  }
   fn setup(srs: (&Vec<G1Affine>,&Vec<G2Affine>),
            model: &Data) ->
-          (Vec<G1Affine>,Vec<G2Affine>);
+          (Vec<G1Affine>,Vec<G2Affine>){
```
ddkang: a formatter would fixed ){ to ) {

src/basic_block.rs
```diff
@@ -38,23 +39,31 @@ impl DataEnc{
...
+    (Vec::new(),Vec::new())
+  }
   fn prove<R: Rng>(srs: (&Vec<G1Affine>,&Vec<G2Affine>),
                    setup: (&Vec<G1Affine>,&Vec<G2Affine>),
                    model: &Data,
                    inputs: &Vec<Data>,
                    output: &Data,
                    rng: &mut R) ->
-                  (Vec<G1Affine>,Vec<G2Affine>,Vec<Fr>);
+                  (Vec<G1Affine>,Vec<G2Affine>){
```
ddkang: and spaces after commas

.gitignore
ddkang: if you use vscode you can have the formatted be automatically applied

arifiorino:
@ddkang I ran rustfmt

ddkang (approved):

src/main.rs
Vimalakirti, line 1: I feel we might want to think about the MSMs happening in the Data::new method:
1. Not all operations are encoded in univariate poly and MSM here. (For MatMult, it is encoded in multilinear poly and do different commitment here)
2. We may want to count the time for MSMs as a part of prover time
3. Since it performs MSMs, I think the name Data::new might be confusing. 

Maybe we can start with Data::new and call Data::encode later to realize what I said above

Vimalakirti (commented):
Comments about Data::new

src/main.rs
arifiorino: For 1, I thought that we had a transformation between a univariate polynomial and a multilinear polynomial. I think the Data object should stay with the univariate polynomial, since every basic block works on univariate polynomials except for matmul. For matmul, we can write this transformation into the basic block itself.
For 2, this repo currently has no timing information so this is irrelevant.
For 3, I think renaming this to Data::encode is a bad idea because there is the DataEnc struct as well, and this would be confusing. It works well the way it is, there is a Data struct which is created from a raw vector, and a DataEnc struct which is created from a Data struct.

Vimalakirti: For 1, agree with the idea, let me think about how to do it.
For 2, I mean maybe the Data::new should be included in the prove method, so that we can count our prover time easily. But right, it is not urgent for now, we can design this later.
For 3, I still think we should use Data::encode instead of including both initialization and MSMs in the Data::new. We can rename the Data and DataEnc as ProverData and VerifierData, it would be less confusing in my opinion.

Vimalakirti (approved):

PR #2, opened 2024-02-04
Simplify cq

arifiorino:
Simplifies the cq protocol by sending `B(x)` instead of `B_0(gamma)`. This makes the protocol much easier to reason about, and easier to convert to zero knowledge in the future.

#### Proof of Soundness
1. `A_x_1` corresponds to the vector `A_i = m_i/(t_i+beta)` by checks 1,2
2. `A_0_1 = A(0)` by check 3
3. `B_x_1` corresponds to the vector `B_i = 1/(f_i+beta)` by checks 4,5,7
4. `B_0_1 = B(0)` by check 6
5. `∑A = A_0_1/N = B_0_1/n = ∑B` by construction of `B_0_1` and cq lemma 2.1
6. `f_i` is contained in `t_i` by cq lemma 2.4

### DIFF
--- a/src/basic_block/cq.rs
+++ b/src/basic_block/cq.rs
@@ -5,9 +5,9 @@ use crate::util;
 use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
 use ark_ec::pairing::Pairing;
 use ark_ff::Field;
-use ark_poly::{evaluations::univariate::Evaluations, univariate::DensePolynomial, EvaluationDomain, GeneralEvaluationDomain, Polynomial};
+use ark_poly::{evaluations::univariate::Evaluations, univariate::DensePolynomial, EvaluationDomain, GeneralEvaluationDomain};
 use ark_std::{
-  ops::{Div, Mul, Sub},
+  ops::{Mul, Sub},
   One, UniformRand, Zero,
 };
 use rand::Rng;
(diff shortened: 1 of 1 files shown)


src/basic_block/cq.rs
```diff
@@ -148,43 +114,49 @@ impl BasicBlock for CQBasicBlock {
     let N = model.len;
     let n = inputs[0].len;
     let domain_n = GeneralEvaluationDomain::<Fr>::new(n).unwrap();
-    let [m_x_1, A_x_1, Q_A_x_1, B_0_x_1, Q_B_x_1, P_x_1, pi_gamma, A_0_x, B_0_gamma_1, b_gamma_1, f_gamma_1, A_0_1, b_gamma_f_gamma] = proof.0[..]
-    else {
+    let [m_x_1, A_x_1, Q_A_x_1, A_0_1, A_0_x_1, B_x_1, Q_B_x_1, B_0_x_1, B_DC] = proof.0[..] else {
```
ddkang, line 117: I'm wondering if it makes sense to have structs for various polys, like A.x A.zero A.one if we're using specific values?

arifiorino: Yes, we could combine the `A` and `B` values into two instances of a unified struct, and then write functions that operate on that struct. This could save some code repetition in the future. But for now I don't think it's necessary since the code is only repeated once.

ddkang: It's more about readability than reptition

PR #3, opened 2024-02-05
zkcq

arifiorino:
#### Changes
- The Data struct now has a parameter `r` that is generated by the prover randomly.
- The DataEnc struct is modified to add `+rY` to the committed value in order to hide it from the verifier. Here `Y` is a group element that is shared by the prover and verifier, but they aren't able to decode it to a field element. I have set `Y = X^{2^28-1}` as it is already in the ptau file and is currently not being used.
- Added zero knowledge to the add, mul, and cq basic blocks by adding a correction factors `C` to the proofs and verifications. Instead of checking `e(a,b) = e(c,d)`, the verifier now checks `e(a,b) = e(c,d) + e(C,Y)` for a correction factor `C` given by the prover.

This pull request builds off the simplify_cq branch which simplified the cq protocol by sending `B(x)` instead of `B_0(gamma)`.

#### cq Soundness Proof
1. `A_x_1` corresponds to the vector `A_i = m_i/(t_i+beta)` by checks 1,2
2. `A_0_1 = A(0)` by check 3
3. `B_x_1` corresponds to the vector `B_i = 1/(f_i+beta)` by checks 4,5,7
4. `B_0_1 = B(0)` by check 6
5. `∑A = A_0_1/N = B_0_1/n = ∑B` by construction of `B_0_1` and cq lemma 2.1
6. `f_i` is contained in `t_i` by cq lemma 2.4

#### cq Zero Knowledge Proof
The simulator will generate a proof that a random vector `a` is in the table `T`. Since the proof includes blinding factors, all values sent to the verifier have `+rY` added to them, and are therefore indistinguishable from random values by a computationally limited adversary.

### DIFF
--- a/src/basic_block.rs
+++ b/src/basic_block.rs
@@ -4,10 +4,11 @@ pub use add::AddBasicBlock;
 use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine};
 use ark_poly::univariate::DensePolynomial;
 use ark_poly::{EvaluationDomain, GeneralEvaluationDomain};
+use ark_std::UniformRand;
 pub use cq::CQBasicBlock;
 pub use cqlin::CQLinBasicBlock;
 pub use mul::MulBasicBlock;
-use rand::Rng;
+use rand::{rngs::StdRng, Rng, SeedableRng};
 pub mod add;
 pub mod cq;
 pub mod cqlin;
@@ -17,6 +18,7 @@ pub struct Data {
   pub raw: Vec<Fr>,
   pub poly: DensePolynomial<Fr>,
   pub g1: G1Affine,
+  pub r: Fr,
 }
 impl Data {
   pub fn new(srs: (&Vec<G1Affine>, &Vec<G2Affine>), raw: &Vec<Fr>) -> Data {
@@ -24,10 +26,12 @@ impl Data {
     let domain = GeneralEvaluationDomain::<Fr>::new(N).unwrap();
     let f = DensePolynomial { coeffs: domain.ifft(raw) };
     let fx: G1Affine = util::msm::<G1Projective>(&srs.0[..N], &f.coeffs).into();
+    let mut rng = StdRng::from_entropy();
     return Data {
       raw: raw.to_vec(),
       poly: f,
       g1: fx,
+      r: Fr::rand(&mut rng),
     };
   }
 }
@@ -36,10 +40,10 @@ pub struct DataEnc {
   pub g1: G1Affine,
 }
 impl DataEnc {
-  pub fn new(data: &Data) -> DataEnc {
+  pub fn new(srs: (&Vec<G1Affine>, &Vec<G2Affine>), data: &Data) -> DataEnc {
     return DataEnc {
       len: data.raw.len(),
-      g1: data.g1,
+      g1: (data.g1 + srs.0[srs.1.len() - 1] * data.r).into(),
     };
   }
 }
--- a/src/basic_block/cq.rs
+++ b/src/basic_block/cq.rs
@@ -5,9 +5,9 @@ use crate::util;
 use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
 use ark_ec::pairing::Pairing;
 use ark_ff::Field;
-use ark_poly::{evaluations::univariate::Evaluations, univariate::DensePolynomial, EvaluationDomain, GeneralEvaluationDomain, Polynomial};
+use ark_poly::{evaluations::univariate::Evaluations, univariate::DensePolynomial, EvaluationDomain, GeneralEvaluationDomain};
 use ark_std::{
-  ops::{Div, Mul, Sub},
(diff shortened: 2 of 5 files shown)


src/basic_block/cq.rs
Vimalakirti, line 1: In my opinion, the blinding part in cq is really difficult to read compared to ones in add and mul. But I think this does achieve zero knowledge, as claimed in Ari's proof.

ddkang: @arifiorino the readability here is very important for having people verify correctness. What do you think about the struct idea? We can either do it now or later

arifiorino: I agree that the code should be made more readable. I think the variable names should be changed. Currently, `A_x_1` stands for `A(x)` encoded as a G1 element, or `[A(x)]_1`.  Since I'm unable to put brackets in a variable name, this got shortened to `A_x_1`. Similarly, `A_0_1 = [A(0)]_1` and `A_0_x_1 = [(A(x)-A(0))/x]_1`. So these mathematical formulas are hard to translate into variable names. I can add a struct for A if it helps readability, but I'm not sure it would be more readable to have `A_x_1` or `A.x_1` instead.

Yes, the blinding part of cq could be made more readable as well. Calculating `C` could potentially be done automatically, as there is a one-to-one correspondence with the pairings in the verify method. But the complicated part of calculating `C` is that if you multiply `(f(x)+r_fY)(g(x)+r_gY)`, there are three correction factors involving randomness that need to be calculated by the prover.

ddkang: We can use . to indicate one thing vs _. In either case, we should think carefully about readability and correctness. What is your proposal?

arifiorino: I added the A,B structs in a new commit

ddkang (approved):
@liliatangxy @Vimalakirti can you look through this PR too?

I don't think it matters for now, but it might be nice if the srs were a struct as well instead of a pair

src/basic_block.rs
```diff
@@ -36,10 +40,10 @@ pub struct DataEnc {
   pub g1: G1Affine,
 }
 impl DataEnc {
-  pub fn new(data: &Data) -> DataEnc {
+  pub fn new(srs: (&Vec<G1Affine>, &Vec<G2Affine>), data: &Data) -> DataEnc {
```
ddkang, line 43: As per comment, I think it would be nicer of the srs were a struct not a pair

src/basic_block/cq.rs
```diff
@@ -5,23 +5,37 @@ use crate::util;
...
+use rand::{rngs::StdRng, Rng, SeedableRng};
 use rayon::prelude::*;
 use std::collections::HashMap;
 
+struct AProof {
+  x: G1Affine,        // A(x)
+  Q_x: G1Affine,      // A(x)T(x) + A(x)beta - m(x) = Q(x)Z(x)
+  zero: G1Affine,     // A(0)
+  zero_div: G1Affine, // (A(x)-A(0))/x
+}
```
ddkang, line 22: generally add new lines after struct defintions

src/basic_block/cq.rs
```diff
@@ -65,77 +79,67 @@ impl BasicBlock for CQBasicBlock {
...
+      -srs.0[1] * r[4] + srs.0[0] * (r[1] - r[3]),
+      -(srs.0[n] - srs.0[0]) * r[6]
+        + inputs[0].g1 * r[5]
+        + B.x * inputs[0].r
+        + (srs.0[srs.1.len() - 1] * inputs[0].r * r[5])
+        + srs.0[0] * (r[5] * beta),
+      -srs.0[1] * r[7] + srs.0[0] * (r[5] - r[3] * Fr::from(N as u32) * Fr::from(n as u32).inverse().unwrap()),
+      -srs.0[0] * r[8] + srs.0[N - n] * r[5],
+    ];
+    let mut C: Vec<G1Affine> = C.iter().map(|x| (*x).into()).collect();
```
ddkang, line 139: What is this doing?

liliatangxy: Can you give names to the blinding factors?

arifiorino: Line 125 generates the random blinding factors `r` that will be used to blind the proof. Line 126 makes the proof vector. Line 127 adds the blinding factors to the proof (add `+rY` to each element of the proof vector). Lines 128-138 calculate the 5 correction factors `C` that are necessary for the verifier to use for their pairing checks. And line 139 converts the correction factors from projective form into affine form.

liliatangxy:
I think the blinding mechanism looks correct. Like Ari suggested, we may want to implement some way to compute the blinding automatically via a wrapper for proof objects. However, at least for cqlin they only reach up to 3 terms.

The structs added work for readability. I don't know of a cleaner way to do the wrapping and unwrapping at the prover and verifier.

liliatangxy (approved):

ddkang:
Go ahead and merge, just a few things to think about

PR #4, opened 2024-02-06
NDArray, MatMult, ZeroMorph, and Bridge stuffs

Vimalakirti:
This branch contains the following features:
- Replace Vec<Vec<Fr>> by Tensor to include shape information and the supports from that ndarray
- MatMult protocol in zkVPD scheme (similar to [Libra](https://eprint.iacr.org/2019/317.pdf)) w/ ZeroMorph
- Bridge argument between ZeroMorph commitment and KZG commitment
- Rope argument (but this might be replaced by integrating multiple existing arguments (i.e., mul, permute, ...))
- Transpose argument (but we might have no need to use this argument; currently, it is not maintained)

(Not ready to be merged. But it is good to read and think about how to integrate these into master branch.)
Still in progress or debugging:
- ZeroMorph crashed for some polynomials, it might be because of my ignorance of some boundary conditions
- Improve readability of the codes
- Append argument (it might also be not necessary)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+debug/
+target/
+
+Cargo.lock
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,5 +9,10 @@ ark-ff = {version = "0.4", features = ["parallel"]}
 ark-ec = "0.4"
 ark-poly = {version = "0.4", features = ["parallel"]}
 ark-bn254 = "0.4"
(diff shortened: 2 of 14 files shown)


ddkang:
Whenever it's ready, can this PR be split into smaller PRs?

Vimalakirti:
> Whenever it's ready, can this PR be split into smaller PRs?

Yea, I will!

PR #5, opened 2024-02-13
add ndarray

Vimalakirti:
This branch replaces Vec<Vec<Fr>> by Tensor<Fr> to include shape information and the supports from crate ndarray

### DIFF
--- a/src/basic_block/cq.rs
+++ b/src/basic_block/cq.rs
@@ -1,22 +1,30 @@
 #![allow(non_snake_case)]
 #![allow(non_upper_case_globals)]
-use super::{BasicBlock, Data, DataEnc};
-use crate::util;
-use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
 use ark_ec::pairing::Pairing;
 use ark_ff::Field;
-use ark_poly::{evaluations::univariate::Evaluations, univariate::DensePolynomial, EvaluationDomain, GeneralEvaluationDomain, Polynomial};
-use ark_std::{
-  ops::{Div, Mul, Sub},
-  One, UniformRand, Zero,
-};
+use ark_poly::{evaluations::univariate::Evaluations, GeneralEvaluationDomain, EvaluationDomain,univariate::DensePolynomial, Polynomial};
+use ark_bn254::{Fr, G1Projective, G2Projective, G1Affine, G2Affine, Bn254};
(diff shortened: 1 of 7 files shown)


Vimalakirti:
btw, there is also API changes for Proof and Setup to make it more flexible.
Some Rustfmt issues can be ignored since it will be fixed in the later commits.

src/basic_block/cq.rs
```diff
@@ -1,22 +1,30 @@
...
 use rand::Rng;
 use rayon::prelude::*;
-use std::collections::HashMap;
+use super::{BasicBlock,Data,DataEnc,Tensor};
+use crate::util;
 
 pub struct CQBasicBlock;
 impl BasicBlock for CQBasicBlock {
-  fn setup(srs: (&Vec<G1Affine>, &Vec<G2Affine>), model: &Data) -> (Vec<G1Affine>, Vec<G2Affine>) {
+  type Proof = (Vec<G1Affine>, Vec<G2Affine>);
```
liliatangxy, line 16: Do we still want to keep the previous proof structs to use with this?

liliatangxy (approved):

src/basic_block/cq.rs
ddkang: What do you mean?

liliatangxy: Ari made some readability changes to reduce operating on unstructured proof vectors: https://github.com/uiuc-kang-lab/zkml_proofs/blob/master/src/basic_block/cq.rs#L17. It would be clearer to continue using this with the proof and setup and change the type to include those structs. This API change is independent of ndarray though so it can be addressed separately.

arifiorino:
First off, there are merging issues because this was branched off before blinding factors were added:
Screenshot 2024-02-15 at 9 56 17 AM
But those issues can be fixed manually.

There are several issues I have found with this PR:
1. rustfmt has not ben run so this is incorrectly formatted.
2. The model is defined as Vec\<Tensor\> when it should be just Tensor.
3. For cqlin, the model should be a 2d tensor, so it should be model[i][j] instead of model[j*n+i]
4. The basic_block trait has not been modified, but the structs that implement that trait have been modified.
5. When I try to compile this code, there are 30 errors and it doesn't compile.

Vimalakirti:
Hi, I patch some compiling errors, so it should now work well.

PR #8, opened 2024-02-15
cqlin rectangle

liliatangxy:
**Changes**
Anything with `L_i` is changed to be over order `m`. The less obvious instances include
- `S` where `1/n` becomes `1/m` in the setup and corresponding check. `V` is divided by `m` instead based on the identity used to calculate `S_i` in Lemma 7.1.
- Domain used to calculate `Q` is now order `2m`.
- `L_Vi(X)` is needed to compute `pi`.
- deg(R) < mn check and `P_R` is added to account for differently-sized cqlins. Though technically I kept `N=m*n` so the check isn't useful.

The API changes will likely be changed back based on ndarray. Some readability changes we discussed will be added on rebase.


arifiorino:
This PR is looking good so far. I think there was a bit of confusion though about `srs.N` from the tensorplonk papar. This `srs.N` is the largest power in the srs, not `m*n`. This number is used for degree checks. Logically, if we want to prove that `deg(f)<n`, we can have the prover generate `f(x) * x^(srs.N-n)`, which they wouldn't be able to do if `deg(f)>=n` since they only have `srs.N` powers of `x`.

PR #9, opened 2024-02-19
Switched to tensors

arifiorino:
The Data and DataEnc structs now represent a N dimensional tensor instead of a vector.
The Add, Mul, and CQ basic blocks now operate on inputs of any dimension.
The CQLin basic block operates on a model with shape n x n and an input vector of shape n.

### DIFF
--- a/src/basic_block.rs
+++ b/src/basic_block.rs
@@ -8,6 +8,7 @@ use ark_std::UniformRand;
 pub use cq::CQBasicBlock;
 pub use cqlin::CQLinBasicBlock;
 pub use mul::MulBasicBlock;
+use ndarray::ArrayD;
 use rand::{rngs::StdRng, Rng, SeedableRng};
 pub mod add;
 pub mod cq;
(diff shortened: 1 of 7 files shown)


src/basic_block.rs
```diff
@@ -8,27 +8,29 @@ use ark_std::UniformRand;
...
   pub poly: DensePolynomial<Fr>,
   pub g1: G1Affine,
   pub r: Fr,
 }
 impl Data {
-  pub fn new(srs: (&Vec<G1Affine>, &Vec<G2Affine>), raw: &Vec<Fr>) -> Data {
-    let N = (*raw).len();
+  pub fn new(srs: (&Vec<G1Affine>, &Vec<G2Affine>), raw: &ArrayD<Fr>) -> Data {
+    let N = raw.len();
+    let vec = raw.clone().into_raw_vec();
```
ddkang, line 27: Does this do a copy?

src/basic_block.rs
```diff
@@ -8,27 +8,29 @@ use ark_std::UniformRand;
...
     let domain = GeneralEvaluationDomain::<Fr>::new(N).unwrap();
-    let f = DensePolynomial { coeffs: domain.ifft(raw) };
+    let f = DensePolynomial { coeffs: domain.ifft(&vec) };
     let fx: G1Affine = util::msm::<G1Projective>(&srs.0[..N], &f.coeffs).into();
     let mut rng = StdRng::from_entropy();
     return Data {
-      raw: raw.to_vec(),
+      raw: raw.clone(),
```
ddkang, line 33: or is this a pointer?

src/basic_block.rs
```diff
@@ -37,19 +39,21 @@ impl Data {
...
       len: data.raw.len(),
+      shape: data.raw.shape().to_vec(),
       g1: (data.g1 + srs.0[srs.1.len() - 1] * data.r).into(),
     };
   }
 }
 pub trait BasicBlock {
-  fn run(model: &Vec<Fr>, inputs: &Vec<Vec<Fr>>) -> Vec<Fr> {
-    Vec::new()
+  fn run(model: &ArrayD<Fr>, inputs: &Vec<ArrayD<Fr>>) -> ArrayD<Fr> {
```
ddkang, line 55: can some basis blocks return multiple ArrayD?

src/basic_block.rs
arifiorino: Yes, the ArrayD needs to be converted to a vector in order to do an fft on it. This requires a clone because converting it destroys the original ArrayD object.

src/basic_block.rs
arifiorino: Yes, I decided to do a copy so that the Data object owns `raw`. I don't think this will be an issue since these are just field elements, which don't take up much space.

src/basic_block.rs
arifiorino: Not at the moment. Right now they have multiple inputs and one output. But this is something that could be added in the future.

ddkang (approved):

PR #10, opened 2024-02-19
cqlin rectangle

liliatangxy:
**Changes**
Anything with `L_i` is changed to be over order `m`. The less obvious instances include
- `S` where `1/n` becomes `1/m` in the setup and corresponding check. `V` is divided by `m` instead based on the identity used to calculate `S_i` in Lemma 7.1.
- Domain used to calculate `Q` is now order `2m`.
- `L_Vi(X)` is needed to compute `pi`.
- deg(R) < mn check and `P_R` is added. For now used `N = srs.1.len() - 1`.


src/basic_block/cqlin.rs
```diff
@@ -6,151 +6,185 @@ use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
...
+use ark_std::{UniformRand, Zero};
 use ndarray::ArrayD;
 use rand::Rng;
 use rayon::prelude::*;
 
 pub struct CQLinBasicBlock;
 impl BasicBlock for CQLinBasicBlock {
   fn run(model: &ArrayD<Fr>, inputs: &Vec<ArrayD<Fr>>) -> ArrayD<Fr> {
-    let n = inputs[0].len();
+    let m = model.shape()[0];
```
ddkang, line 18: Is model the right name here?

ddkang: Maybe we can solve it later

src/basic_block/cqlin.rs
```diff
@@ -6,151 +6,185 @@ use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
...
     }
     return r;
   }
   fn setup(srs: (&Vec<G1Affine>, &Vec<G2Affine>), model: &Data) -> (Vec<G1Affine>, Vec<G2Affine>) {
-    let N = model.raw.len();
-    let n: usize = (N as f64).sqrt() as usize;
-    let n_inv = Fr::from(n as u64).inverse().unwrap();
+    let m = model.raw.shape()[0];
+    let n = model.raw.shape()[1];
+    let N = srs.1.len() - 1;
```
ddkang, line 31: Can we used a named tuple for the srs? I think srs.g1 and srs.g2 would be more clear

src/basic_block/cqlin.rs
```diff
@@ -6,151 +6,185 @@ use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
...
       .collect();
     let S: Vec<_> = S.par_iter().map(|x| x.iter().sum::<G1Projective>()).collect();
-    let R: Vec<Vec<_>> = (0..n).into_par_iter().map(|i| (0..n).map(|j| U[i][j] * model.raw[[i, j]]).collect()).collect();
+
+    let R: Vec<Vec<_>> = (0..m).into_par_iter().map(|i| (0..n).map(|j| U[i][j] * model.raw[[i, j]]).collect()).collect();
     let R: Vec<_> = R.par_iter().map(|x| x.iter().sum::<G1Projective>()).collect();
 
-    let mut C: Vec<Vec<_>> = (0..n).into_par_iter().map(|i| (0..n).map(|j| model.raw[[j, i]]).collect()).collect();
-    C.par_iter_mut().for_each(|x| domain_n.ifft_in_place(x));
+    let P_R: Vec<Vec<_>> = (0..m).into_par_iter().map(|i| (0..n).map(|j| U_P_R[i][j] * model.raw[[i, j]]).collect()).collect();
```
ddkang, line 82: Why do we collect this twice?

src/basic_block/cqlin.rs
```diff
@@ -6,151 +6,185 @@ use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
...
+    let P_R: Vec<_> = P_R.par_iter().map(|x| x.iter().sum::<G1Projective>()).collect();
+
+    let mut C: Vec<Vec<_>> = (0..n).into_par_iter().map(|i| (0..m).map(|j| model.raw[[j, i]]).collect()).collect();
+    C.par_iter_mut().for_each(|x| domain_m.ifft_in_place(x));
 
     let mut temp = C;
```
ddkang: Regardless of whether this is a reference or copy, I think it would be better to make it obvious

src/basic_block/cqlin.rs
liliatangxy: I guess they should be equal. There is iteration of 0..m over both model and input.

ddkang:
There's conflicts with main it seems

src/basic_block/cqlin.rs
liliatangxy: Can fix that in next PR over all files

src/basic_block/cqlin.rs
ddkang: Should we assert that they are equal?

liliatangxy:
Um I actually merged it already since I thought the code was already reviewed last time. The conflict is already fixed. I will fix the above over all files in another PR if relevant.

ddkang:
Alright, fix it in another PR

src/basic_block/cqlin.rs
liliatangxy: It is for each of the nested iterators.

ddkang: Can we benchmark this for large matrixes? I suspect nested parallel iterators without collection is more efficient since Rust can handle the scheduling

arifiorino: Right now in the code we have several variables that are `Vec<Vec<_>>` as in the original cqlin code. Eventually we could move these to 2D ndarrays to make it easier to read and more efficient.

src/basic_block/cqlin.rs
```diff
@@ -6,151 +6,185 @@ use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
...
-    util::ifft_in_place(domain_n, &mut L_i_x);
-    let mut L_i_x_n: Vec<_> = (0..n).into_par_iter().map(|i| srs_p[n * i]).collect();
-    util::ifft_in_place(domain_n, &mut L_i_x_n);
-
-    let mut temp: Vec<Vec<_>> = (0..n).into_par_iter().map(|i| (0..n).map(|j| srs_p[i + n * j]).collect()).collect();
-    temp.par_iter_mut().for_each(|x| util::ifft_in_place(domain_n, x));
-    let mut U: Vec<Vec<_>> = (0..n).into_par_iter().map(|j| (0..n).map(|i| temp[i][j]).collect()).collect();
+    let domain_2m = GeneralEvaluationDomain::<Fr>::new(2 * m).unwrap();
+
+    let srs_p: Vec<G1Projective> = srs.0[..N].into_par_iter().map(|x| (*x).into()).collect();
```
arifiorino: So this converts `2^28` affine points to projective points. The resulting vector will be `96 * 2^28` bytes which is 26GB. The issue is that most of this vector will not be used because we only use the first `n*m` and the last `n*m` elements of this vector. So I think we should only set `srs_p` to be the first `n*m` elements, and handle the `U_P_R` case seperately

src/basic_block/cqlin.rs
```diff
@@ -163,4 +197,4 @@ impl BasicBlock for CQLinBasicBlock {
     let rhs = Bn254::pairing(pi_1, srs.1[n]);
     assert!(lhs == rhs);
   }
-}
+}
```
arifiorino: Very minor thing, but I was assuming our files end in newlines. Not sure if there is an IDE setting for this.

src/main.rs
```diff
@@ -29,10 +29,11 @@ fn main() {
...
+  const m: usize = 1 << 2;
   let a: Vec<_> = (0..N).into_par_iter().map_init(rand::thread_rng, |rng, _| Fr::rand(rng)).collect();
   let b: Vec<_> = (0..N).into_par_iter().map_init(rand::thread_rng, |rng, _| Fr::rand(rng)).collect();
   test_basic_block::<AddBasicBlock>(srs, &arr1(&vec![]).into_dyn(), &vec![arr1(&a).into_dyn(), arr1(&b).into_dyn()]);
   test_basic_block::<MulBasicBlock>(srs, &arr1(&vec![]).into_dyn(), &vec![arr1(&a).into_dyn(), arr1(&b).into_dyn()]);
   test_basic_block::<CQBasicBlock>(srs, &arr1(&a).into_dyn(), &vec![arr1(&a[..n]).into_dyn()]);
-  test_basic_block::<CQLinBasicBlock>(srs, &ArrayD::from_shape_vec(vec![n, n], a).unwrap(), &vec![arr1(&b[..n]).into_dyn()]);
-}
+  test_basic_block::<CQLinBasicBlock>(srs, &ArrayD::from_shape_vec(vec![m, N/m], a).unwrap(), &vec![arr1(&b[..m]).into_dyn()]);
+}
```
arifiorino: Just a reminder to run `cargo fmt` before you commit. Cause `N/m` should be `N / m`.

src/basic_block/cqlin.rs
liliatangxy: Ran each of these 20 times over 16 cores on random cc machine, m = 2^3, n = 2^9, opt=3

current one
```rust
let P_R: Vec<Vec<_>> = (0..m).into_par_iter().map(|i| (0..n).map(|j| U_P_R[i][j] * model.raw[[i, j]]).collect()).collect();
let P_R: Vec<_> = P_R.par_iter().map(|x| x.iter().sum::<G1Projective>()).collect();
```

combined without inner collect, parallel over m
```rust
let P_R: Vec<_> = (0..m).into_par_iter().map(|i| (0..n).map(|j| U_P_R[i][j] * model.raw[[i, j]]).sum::<G1Projective>()).collect();
```

combined without inner collect, parallel over n
```rust
let P_R: Vec<_> = (0..m).map(|i| (0..n).into_par_iter().map(|j| U_P_R[i][j] * model.raw[[i, j]]).sum::<G1Projective>()).collect();
```
Total (including 20 runs) for each is about 14 s. The middle one is consistently fastest. Over different runs the difference between them is about 1-4 ms avg for an individual run. Same is true for m = n = 2^6.

Will fix

ddkang: I don't think I understand. What is the mean runtime for 1, 2, and 3?

liliatangxy: 1: 0.703195 ms
2: 0.701535
3: 0.7061595

ddkang: I think you want to try a much larger matrix, like the size used in Llama

liliatangxy: For m=2^8, n=2^19 on ddkang machine with 32 cores
1: 11.146811988s
2: 11.855352996s
3: 11.09063273s

ddkang: hmm okay I don't think it really matters in that case

src/basic_block/cqlin.rs
liliatangxy: It is moved. How should we indicate this?

ddkang: wait if it's moved why do we need this?

liliatangxy: I guess it was there for readability since the final `let C` corresponds to what is in the paper and temp following modifies it in place. I can add some comments that can let us get rid of this line.

arifiorino (approved):

src/main.rs
ddkang: you can add this as a VSCode setting

src/basic_block/cqlin.rs
ddkang: has this actually been fixed?

liliatangxy: yes

PR #11, opened 2024-02-22
zk cqlin

liliatangxy:
**Changes**
Add zero knowledge via blinding over all proof objects in cqlin with `+rY` in the same way as #3.

All blinding term calculations are done similarly as in zkcq.

### DIFF
--- a/src/basic_block/cqlin.rs
+++ b/src/basic_block/cqlin.rs
@@ -162,7 +162,7 @@ impl BasicBlock for CQLinBasicBlock {
     proof.push(z);
 
     // G1 M needed for blinding
-    // h, h_M, h_S, h_g, h_R, h_pi, h_pi_1
+    // h, h_S, h_g, h_R, h_pi, h_pi_1
     let M_x_1 = R.iter().sum::<G1Projective>();
     let C = vec![
       M_x_1 * r[2] - (srs.0[m * n] - srs.0[0]) * r[1] - srs.0[0] * r[0] + srs.0[srs.1.len() - 1] * r[2] * r[8],
--- a/src/basic_block/cqlin.rs
+++ b/src/basic_block/cqlin.rs
@@ -8,7 +8,7 @@ use ark_ff::Field;
 use ark_poly::{EvaluationDomain, GeneralEvaluationDomain, Polynomial};
 use ark_std::{UniformRand, Zero};
 use ndarray::ArrayD;
-use rand::Rng;
+use rand::{rngs::StdRng, Rng, SeedableRng};
 use rayon::prelude::*;
 
 pub struct CQLinBasicBlock;
@@ -140,7 +140,7 @@ impl BasicBlock for CQLinBasicBlock {
     let v = inputs[0].raw.clone().into_raw_vec();
     let R_x = util::msm::<G1Projective>(R, &v).into();
     let Q_x = util::msm::<G1Projective>(Q, &v).into();
-    let temp: Vec<_> = (0..n).into_par_iter().map(|i| srs.0[n * i]).collect();
+    let temp: Vec<_> = (0..m).into_par_iter().map(|i| srs.0[n * i]).collect();
     let A_x = util::msm::<G1Projective>(&temp, &inputs[0].poly.coeffs).into();
     let S_x = util::msm::<G1Projective>(S, &v).into();
     let P_x = util::msm::<G1Projective>(&srs.0[N - n..N], &output.poly.coeffs).into();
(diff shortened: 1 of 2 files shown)


src/basic_block/cqlin.rs
```diff
@@ -154,11 +154,35 @@ impl BasicBlock for CQLinBasicBlock {
...
     let pi_1 = util::msm::<G1Projective>(&L_V_i_x_n, &h_i).into();
 
-    // TODO: Implement blinding
-    let C0 = (srs.0[srs.1.len() - 1] * inputs[0].r).into();
-    let C1 = (srs.0[srs.1.len() - 1] * output.r).into();
-
-    return (vec![R_x, Q_x, A_x, S_x, P_x, P_R_x, pi, pi_1, z, C0, C1], vec![setup.1[0]]);
+    let mut rng2 = StdRng::from_entropy();
+    // R, Q, A, S, P, pR, pi, pi_1, M
+    let r: Vec<_> = (0..9).map(|_| Fr::rand(&mut rng2)).collect();
```
ddkang, line 159: does it matter how the r is computed? as in, can the prover cheat by choosing a bad r?

src/basic_block/cqlin.rs
```diff
@@ -140,7 +140,7 @@ impl BasicBlock for CQLinBasicBlock {
     let v = inputs[0].raw.clone().into_raw_vec();
     let R_x = util::msm::<G1Projective>(R, &v).into();
     let Q_x = util::msm::<G1Projective>(Q, &v).into();
-    let temp: Vec<_> = (0..n).into_par_iter().map(|i| srs.0[n * i]).collect();
+    let temp: Vec<_> = (0..m).into_par_iter().map(|i| srs.0[n * i]).collect();
```
arifiorino, line 143: It's good that this was fixed cause cqlin was broken when `m>n`. We should have caught this, maybe we should make some sample tests in the future.

src/basic_block/cqlin.rs
```diff
@@ -154,11 +154,35 @@ impl BasicBlock for CQLinBasicBlock {
...
+      srs.0[0] * inputs[0].r + (srs.0[0] * gamma_n - srs.0[1]) * r[6],
+      srs.0[0] * r[2] + (srs.0[0] * gamma_n - srs.0[n]) * r[7],
+    ];
+
+    let mut C: Vec<G1Affine> = C.iter().map(|x| (*x).into()).collect();
+    proof.append(&mut C);
+
+    // G2 blinding for M
+    let M_x_2 = (setup.1[0] + srs.1[srs.1.len() - 1] * r[8]).into();
+    let temp: Vec<_> = (0..m).into_par_iter().map(|i| srs.1[n * i]).collect();
```
arifiorino: We don't need to do this. All we need to do is add `+ A_x * r[8]` to `C1` and we're good. Here is a patch for this: [C1_fix.patch](https://github.com/uiuc-kang-lab/zkml_proofs/files/14380980/C1_fix.patch)

arifiorino (changes requested):

src/basic_block/cqlin.rs
liliatangxy: Oh yeah you are right

src/basic_block/cqlin.rs
arifiorino: I think the prover can theoretically cheat since `Y=X^(2^28-1)`. We should set `Y` randomly in the future. One potential idea is using the challenge_84 file for `Y` and the challenge_85 file for `X`.

src/basic_block/cqlin.rs
ddkang: did you push the fix?

src/basic_block/cqlin.rs
ddkang: can we fix this now?

liliatangxy: I'm not sure if Y is an issue since the verifier also knows Y. The `gamma` opening argument in round 2 should be generated by Fiat-Shamir.

src/basic_block/cqlin.rs
liliatangxy: fixed now

src/basic_block/cqlin.rs
arifiorino: The issue is if the prover could somehow obfuscate their commitments using the fact that `Y=X^(2^28-1)`. This should be an easy fix, and I was going to include it in my next PR. It might be easier to merge this PR first though.

arifiorino: And yeah the `rng` in the prove function is going to be replaced by a fiat-shamir hash function.

arifiorino (approved):

PR #12, opened 2024-02-23
Graph

arifiorino:
Changes:
- Add `ConstBasicBlock` and `ReLUBasicBlock`
- Made `BasicBlock` object-safe
  - Added `&self` parameters
  - Removed generic parameters: `R` becomes `StdRng`
- Switched from passing a vector of objects to passing a vector of pointers, which avoids copying data
- Made `Graph` struct
  - `run` is a DFS
  - `setup` is in parallel over the blocks
  - `prove` and `verify` are in parallel over the nodes
- Proved a sample `graph`: a 1-layer NN with Linear and ReLU layers
- Moved `BasicBlock` tests to a unit test (invoked with `cargo test`)

### DIFF
--- a/src/basic_block.rs
+++ b/src/basic_block.rs
@@ -67,7 +67,7 @@ pub trait BasicBlock {
     srs: (&Vec<G1Affine>, &Vec<G2Affine>),
     setup: (&Vec<G1Affine>, &Vec<G2Affine>),
     model: &Data,
-    inputs: &Vec<Data>,
+    inputs: &Vec<&Data>,
     output: &Data,
     rng: &mut StdRng,
   ) -> (Vec<G1Affine>, Vec<G2Affine>) {
@@ -77,7 +77,7 @@ pub trait BasicBlock {
     &self,
     srs: (&Vec<G1Affine>, &Vec<G2Affine>),
     model: &DataEnc,
-    inputs: &Vec<DataEnc>,
+    inputs: &Vec<&DataEnc>,
     output: &DataEnc,
     proof: (&Vec<G1Affine>, &Vec<G2Affine>),
     rng: &mut StdRng,
--- a/src/basic_block.rs
+++ b/src/basic_block.rs
@@ -5,6 +5,7 @@ use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine};
 use ark_poly::univariate::DensePolynomial;
 use ark_poly::{EvaluationDomain, GeneralEvaluationDomain};
 use ark_std::UniformRand;
+pub use constant::ConstBasicBlock;
 pub use cq::CQBasicBlock;
 pub use cqlin::CQLinBasicBlock;
 pub use mul::MulBasicBlock;
@@ -12,6 +13,7 @@ use ndarray::ArrayD;
 use rand::{rngs::StdRng, SeedableRng};
 pub use relu::ReLUBasicBlock;
 pub mod add;
+pub mod constant;
 pub mod cq;
 pub mod cqlin;
 pub mod mul;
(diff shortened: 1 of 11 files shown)


src/basic_block.rs
```diff
@@ -52,29 +56,31 @@ impl DataEnc {
...
+  fn prove(
+    &self,
     srs: (&Vec<G1Affine>, &Vec<G2Affine>),
     setup: (&Vec<G1Affine>, &Vec<G2Affine>),
     model: &Data,
-    inputs: &Vec<Data>,
+    inputs: &Vec<&Data>,
     output: &Data,
-    rng: &mut R,
+    rng: &mut StdRng,
```
ddkang, line 72: Could you explain the reasoning behind changing it to a fixed type of random?

src/graph.rs
```diff
@@ -0,0 +1,96 @@
...
+impl Graph {
+  pub fn run(&self, inputs: &Vec<&ArrayD<Fr>>, models: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
+    let mut outputs = vec![ArrayD::zeros(vec![]); self.nodes.len()];
+    // Run the nodes that have no inputs
+    for i in 0..self.nodes.len() {
+      if self.nodes[i].input_nodes.len() == 0 && i != self.input_node {
+        outputs[i] = self.basic_blocks[self.nodes[i].basic_block].run(&models[self.nodes[i].basic_block], &vec![]);
+      }
+    }
+    // DFS:
```
ddkang, line 28: Also could you explain the reason behind doing DFS vs BFS?

src/basic_block.rs
arifiorino: In order to make the `Graph` struct, it has to be able to store `BasicBlock` objects. In order to create these objects, the `BasicBlock` trait has to be "object-safe", which has two requirements: every function has the `&self` parameter, and there are no generic arguments to the functions. As the code was written, the function `prove<R: Rng>(..., rng: R)` had a generic argument `R` that implemented the `Rng` trait. But `R` had an unknown amount of space required to allocate it, so the `BasicBlock` struct had an unknown amount of space required to allocate it, and therefore `BasicBlock` was not object-safe.

src/graph.rs
arifiorino: There is no reason, if I changed the stack to a queue it would be a BFS, it works either way.

ddkang: Are there any performance considerations either way?

src/basic_block.rs
ddkang: I don't know if this makes sense since I'm not sure if StdRng is cryptographically secure... can you check on that?

arifiorino: I checked and `StdRng` is cryptographically secure: [https://rust-random.github.io/book/guide-rngs.html#cryptographically-secure-pseudo-random-number-generators-csprngs](https://rust-random.github.io/book/guide-rngs.html#cryptographically-secure-pseudo-random-number-generators-csprngs)

But eventually this will be replaced by a `Rng` that we create in order to implement Fiat-Shamir.

src/graph.rs
arifiorino: No, it's `O(n)` either way. `n` is the number of nodes in the graph.

liliatangxy: I agree that there aren't other things to consider. This DFS is only for running the model to get the basic block inputs and outputs which happens before proving, so it also isn't very important.

src/basic_block.rs
ddkang: Let's open an issue

arifiorino: Ok, I opened an issue for this.

liliatangxy (approved):

ddkang (approved):

Vimalakirti:
I have some questions about this branch (excuse me for the late feedback since I was doing ZKML stuffs few days ago):
- Why do we have no need to implement `prove()` and `verify()` for `ReLUBasicBlock`?
- What's the purpose of `ConstBasicBlock`?
- We currently run `Data:new()` for all inputs and outputs in `main.rs`, does that means we perform redundant commitment for some data? I guess the answer is yes since some outputs from one basic block are exactly the inputs of other basic blocks. I suggest we can modify this part so that we can save some proving time by preventing redundant MSMs.

arifiorino:
Answers to your questions:
- Because `ReLUBasicBlock` only calculates the ReLU, after this, a cq check does the proof/verification process
- It's just a basic block that is a fixed number that the prover and verifier both know.
- No, we only run `Data::new()` once for each output of the basic blocks, so there is no wasted work.

Vimalakirti:
I see, thanks for the prompt response. I have no question for the PR now.

PR #15, opened 2024-03-21
LLaMA 2-7B Components

arifiorino:
- New struct `SRS` which has X and Y elements in G1 and G2, and in both affine and projective form
- `run`, `setup`, `prove`, `verify` all operate on a vector of commitments
- `setup` and `prove` have affine points as input and projective points as output
- New BasicBlocks:
  - `CQ2`: Batched CQ which takes in two tables and two input vectors
  - `MatMul`: Takes row commitments of `A` and column commitments of `B` and outputs row commitments of `C=AB`
  - `MulScalar` and `MulConst` because multiplying by an unknown scalar or a known constant is much faster than a full multiplication argument
  - `Add`/`Subtract` now allow adding a scalar to a vector
  - `Transpose`: Transposes a matrix
  - `Sum`: Sums all the values in a matrix
  - `Concat`: Concatenates the rows of a matrix into a single vector
  - `Alternating`: Checks that a vector `A` alternating with a vector `B` equals a vector `C`
  - `Eq`: Checks that two vectors are equal
- Other new BasicBlocks proved via `CQ2`:
  - `DivConst`, `Exp`, `Log`, `Sqrt`
- New basic blocks proved by other means:
  - `Combine` combines two vectors into a single one by alternating between the two
  - `Split` splits a vector in two by taking odd and even indices
  - `DivScalar` divides a vector by a scalar and produces both the quotient and remainder vectors
  - `RoPE` returns the constants used in the RoPE operation
- `CQ` and `CQ2` save their internal `HashMap` so it doesn't have to be recalculated in every proof
- Created tests in `tests.rs` for all new BasicBlocks that need them

**New 4/1/2024:**
- Added `BatchedBasicBlock` which runs on `NDArray`
- Switched `Graph` to run on `BatchedBasicBlock`
- `BatchedBasicBlock` checks the dimensions of the `model` and `input` before `run`/`setup`/`prove`/`verify` methods


ddkang:
This PR is quite difficult to review, let's try to break things up in the future

Vimalakirti:
Haven't read all the codes yet, but here's one quick review:
I notice that we no longer use "ArrayD" after this PR since we perform row-wise proving for all operations now; however, it also means we need to do additional works to handle shape info when creating graph.

src/graph.rs
```diff
@@ -1,81 +1,57 @@
...
-        let myInputs = currNode.input_nodes.iter().map(|i| &(outputs[*i])).collect();
-        outputs[curr] = self.basic_blocks[currNode.basic_block].run(&models[currNode.basic_block], &myInputs);
-      }
-      for n in &currNode.output_nodes {
-        if *self.nodes[*n].input_nodes.last().unwrap() == curr {
-          stack.push(*n);
-        }
-      }
-    }
+  pub fn run(&self, inputs: &Vec<&Vec<Fr>>, models: &Vec<&Vec<&Vec<Fr>>>) -> Vec<Vec<Vec<Fr>>> {
```
ddkang: This seems really ugly, I don't understand why we aren't using ArrayD

ddkang: As I've mentioned several times, this prevents us from enforcing any shape constraints on the vectors

ddkang: If this is externally called (via another library) this can lead to unexpected or unsafe behavior

src/basic_block/concat.rs
```diff
@@ -0,0 +1,172 @@
...
+use super::{BasicBlock, Data, DataEnc, SRS};
+use crate::util;
+use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
+use ark_ec::pairing::Pairing;
+use ark_ff::Field;
+use ark_poly::{univariate::DensePolynomial, EvaluationDomain, GeneralEvaluationDomain};
+use ark_std::{ops::Mul, ops::Sub, One, UniformRand, Zero};
+use rand::{rngs::StdRng, SeedableRng};
+
+fn calc_pow(alpha: Fr, n: usize) -> Vec<Fr> {
```
Vimalakirti, line 12: This function duplicates everywhere. I suggest we may move it to `util.rs`.

src/basic_block/eq.rs
```diff
@@ -0,0 +1,68 @@
...
+    proof: (&Vec<G1Affine>, &Vec<G2Affine>),
+    _rng: &mut StdRng,
+  ) {
+    // Verify f(x)+g(x)=h(x)
+    let lhs = Bn254::pairing(inputs[0].g1, srs.X2A[0]);
+    let rhs = Bn254::pairing(inputs[1].g1, srs.X2A[0]) + Bn254::pairing(proof.0[0], srs.Y2A);
+    assert!(lhs == rhs);
+  }
+}
+pub struct AddBasicBlock;
```
Vimalakirti: Why do we need this? We already have `AddBasicBlock` (same struct name) in `add.rs`.

src/basic_block/matmul.rs
```diff
@@ -0,0 +1,198 @@
...
+impl BasicBlock for MatMulBasicBlock {
+  fn run(&self, _model: &Vec<&Vec<Fr>>, inputs: &Vec<&Vec<Fr>>) -> Vec<Vec<Fr>> {
+    let l = self.l;
+    let m = inputs[0].len();
+    let n = inputs.len() - l;
+    let mut r = vec![vec![Fr::zero(); n]; l];
+    for i in 0..l {
+      for j in 0..n {
+        for k in 0..m {
+          r[i][j] += inputs[i][k] * inputs[l + j][k];
```
Vimalakirti, line 36: I know `l` is the boundary to separate elements in A and those in B for matmult A*B. However, I personally find it hard to maintain the code. Do we have better way(s) to realize this?

src/basic_block/max.rs
```diff
@@ -0,0 +1,18 @@
...
+use ark_bn254::Fr;
+use ark_std::Zero;
+
+pub struct MaxBasicBlock;
+impl BasicBlock for MaxBasicBlock {
+  fn run(&self, _model: &Vec<&Vec<Fr>>, inputs: &Vec<&Vec<Fr>>) -> Vec<Vec<Fr>> {
+    let mut max = Fr::zero();
+    for i in 0..inputs.len() {
+      for j in 0..inputs[0].len() {
+        if inputs[i][j] < Fr::from(1 << 28) && inputs[i][j] > max {
```
Vimalakirti, line 14: I know `1<<28` is the boundary for us to separate positive and negative numbers. I suggest we can make it a const and put it in `util.rs`

src/tests.rs
```diff
@@ -2,61 +2,92 @@ use crate::basic_block::*;
...
   let inputs = inputs.iter().map(|x| x).collect();
-  let output = DataEnc::new(srs, &output);
-  basic_block.verify(srs, &model, &inputs, &output, (&(proof.0), &(proof.1)), &mut rng2);
+  let outputs: Vec<_> = outputs.iter().map(|x| DataEnc::new(srs, x)).collect();
+  let outputs = outputs.iter().map(|x| x).collect();
+  basic_block.verify(srs, &model, &inputs, &outputs, proof, &mut rng2);
 }
 
 #[test]
 fn testBasicBlocks() {
```
Vimalakirti, line 41: I've run everything here and nothing went wrong. However, I suggest we can start to think about how to do benchmark here or under `bench/`

src/basic_block/alternating.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let alpha_pow = calc_pow(alpha, n); //r is ignored
+    let alpha_pow_coeff = domain_n.ifft(&alpha_pow);
+    let alpha_pow_g2: G2Affine = util::msm::<G2Projective>(&srs.X2A, &alpha_pow_coeff).into();
+    let beta_pow = calc_pow(beta, n); //r is ignored
+    let beta_pow_coeff = domain_n.ifft(&beta_pow);
+    let beta_pow_g2: G2Affine = util::msm::<G2Projective>(&srs.X2A, &beta_pow_coeff).into();
+    let alpha_beta_pow: Vec<Fr> = (0..2 * n).map(|i| if i % 2 == 0 { alpha_pow[i / 2] } else { beta_pow[i / 2] }).collect();
+    let alpha_beta_pow_coeff = domain_2n.ifft(&alpha_beta_pow);
+    let alpha_beta_pow_g2: G2Affine = util::msm::<G2Projective>(&srs.X2A, &alpha_beta_pow_coeff).into();
```
Vimalakirti, line 128: These msms are really expensive for verifier, but fortunately, they can precompute these. Maybe we can design a setup function for verifier (?)

src/tests.rs
Vimalakirti: Also, I think it is good to showcase common ML operations here (like `ReLUBasicBlock`, `RoPEBasicBlock`, and `DivScalarBasicBlock`

src/graph.rs
```diff
@@ -1,81 +1,57 @@
...
-use ark_bn254::{Fr, G1Affine, G2Affine};
-use ndarray::ArrayD;
+use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
 use rand::rngs::StdRng;
 
 pub struct Node {
   pub basic_block: usize,
-  pub input_nodes: Vec<usize>,
-  pub output_nodes: Vec<usize>,
+  pub inputs: Vec<(i32, usize)>, //(node, output #)
```
Vimalakirti, line 9: It may be good to put `models` (i.e., the weight) as one of the attributes for the corresponding Node (?)

src/main.rs
```diff
@@ -14,89 +15,78 @@ mod tests;
...
-  let input = arr1(&input).into_dyn();
-  let inputs = vec![&input];
-  let empty = ArrayD::zeros(vec![]);
-  let constant = arr1(&vec![Fr::from(1 << 6); 1 << 2]).into_dyn();
-  let relu_cq_table = util::gen_cq_table(&graph.basic_blocks[1], 1 << 6);
-  let models = vec![&matrix, &empty, &constant, &empty, &empty, &relu_cq_table];
+  let mut inputs = vec![&input, &input2];
+  inputs.append(&mut matrix);
+  let empty = vec![];
+  let (id, relu_cq_table) = util::gen_cq_table(vec![&graph.basic_blocks[1]], -(1 << 5), 1 << 6);
```
Vimalakirti: We need this after the `BasicBlock` of every nonlinear function (sigmoid, tanh). It might be good to write a helper function later

src/basic_block/alternating.rs
```diff
@@ -0,0 +1,162 @@
...
+    pow[i + 1] = pow[i] * alpha;
+  }
+  pow
+}
+
+pub struct AlternatingBasicBlock;
+// Inputs are X,Y,Z where X alternating with Y equals Z
+// X,Y,Z are multiplied by alpha,beta,alpha_beta to get A,B,C
+impl BasicBlock for AlternatingBasicBlock {
+  fn prove(
```
Vimalakirti, line 28: Clear, I can follow the proving procedure. But we really need a soundness proof for this

src/basic_block/transpose.rs
```diff
@@ -0,0 +1,171 @@
...
+    let n = inputs.len();
+    let mut r = vec![vec![Fr::zero(); n]; m];
+    for i in 0..n {
+      for j in 0..m {
+        r[j][i] = inputs[i][j];
+      }
+    }
+    r
+  }
+  fn prove(
```
Vimalakirti, line 37: - soundness proof required
- also, maybe we can compare the performance of this block with that of Plonk Permutation later

Vimalakirti (approved):
Here is a summary of this review:
- Need to handle shape information later. (the current PR do row-wise provings)
- Some new BasicBlock proving seems to be true (and understandable), but we really need soundness proofs
- Some functions & values duplicate, we can put them in `util.rs` for better maintenance of codes
- Maybe we'll need "verifier_setup" to precompute some msm in verifier's end
- Consider to add formal bench scripts from now on
- It might be good to add examples to showcase how to leverage ML basicblocks (like DivScalar and RoPE)

src/basic_block/alternating.rs
liliatangxy: Wouldn't we still have to count them in the timing? 
We also may be able to use the same alpha and beta across blocks, so we wouldn't have to redo this computation for each block.

src/basic_block/eq.rs
arifiorino: Yes, this was a mistake, I deleted it.

src/basic_block/alternating.rs
arifiorino: I can formalize this, but it will be very similar to the completeness and soundness proofs in the Aurora paper (page 27): https://eprint.iacr.org/2018/828.pdf

src/basic_block/transpose.rs
arifiorino: For the proof, it will be very similar to the completeness and soundness proofs in the Aurora paper (page 27): https://eprint.iacr.org/2018/828.pdf

src/basic_block/matmul.rs
arifiorino: The three parameters to matmul are `l`,`m`,`n`. We have that `A` is `lxm`, `B` is `mxn`, and `C` is `lxn`. (We assume `AB=C`).
The inputs to `MatMulBasicBlock` are the row commitments to `A` (there are `l` of them), and the column commitments to `B` (there are `n` of them). We don't need to have all `l`,`m`,`n` as parameters of `MatMulBasicBlock`, we only need `l` because `m` and `n` can be inferred using `l`.

src/basic_block/cq2.rs
```diff
@@ -0,0 +1,209 @@
...
+  One, UniformRand, Zero,
+};
+use rand::{rngs::StdRng, SeedableRng};
+use rayon::prelude::*;
+use std::collections::HashMap;
+
+pub struct CQ2BasicBlock {
+  pub table_dict: HashMap<(Fr, Fr), usize>,
+}
+impl BasicBlock for CQ2BasicBlock {
```
liliatangxy, line 20: How is CQ2 used in the graph and which blocks are using it?

src/basic_block/matmul.rs
liliatangxy: Is there any performance benefit to combining the A and B row commitments vs providing them as separate inputs? Although it isn't difficult to figure out, it hurts readability somewhat

src/basic_block/alternating.rs
liliatangxy, line 28: Is there a simple way to abstract parts of the aurora prover and verifier? We are currently repeating much of the same proof construction and verifier logic for most of the blocks.

src/tests.rs
liliatangxy: Probably for any block that doesn't have its own prove/verify block.

liliatangxy (approved):
I didn't see any issues with the repurposing of aurora for the different blocks. In addition to Bing-Jyue's comments we want to consider
- whether parts of the provers and verifiers that are doing different types of aurora can be partly abstracted, since much of the logic is the same
- whether we can use the same alpha/beta across different blocks to save msms (probably in a different pr)
- sharing the llama graph itself, so that we can clearly see the construction of the graph and how all the blocks are used. The blocks without their own prover/verifier implemented are unclear in how they are being proved

arifiorino:
**New 4/1/2024:**
- Added `BatchedBasicBlock` which runs on `NDArray`
- Switched `Graph` to run on `BatchedBasicBlock`
- `BatchedBasicBlock` checks the dimensions of `model` and `input` before `run`/`setup`/`prove`/`verify` methods

src/basic_block.rs
```diff
@@ -41,47 +80,155 @@ impl Data {
...
+impl BatchedBasicBlock {
+  pub fn run(&self, model: &Vec<&ArrayD<Fr>>, inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
+    assert_eq!(
+      (model.iter().map(|x| x.ndim()).collect(), inputs.iter().map(|x| x.ndim()).collect()),
+      self.basic_block.get_dims()
+    );
+    let model = arr_to_vec(model);
+    let model = model.iter().map(|x| x).collect();
+    let inputs = arr_to_vec(inputs);
+    let inputs = inputs.iter().map(|x| x).collect();
```
liliatangxy: Is there a reason to use Vec for BasicBlock operations instead of ArrayD?

src/basic_block.rs
```diff
@@ -41,47 +80,155 @@ impl Data {
...
+}
+pub struct BatchedBasicBlock {
+  pub basic_block: Box<dyn BasicBlock>,
+}
+impl BatchedBasicBlock {
+  pub fn run(&self, model: &Vec<&ArrayD<Fr>>, inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
+    assert_eq!(
+      (model.iter().map(|x| x.ndim()).collect(), inputs.iter().map(|x| x.ndim()).collect()),
+      self.basic_block.get_dims()
+    );
```
liliatangxy: Is the dimension assertion + ArrayD the only purpose for the `BatchedBasicBlock`? Would it be better to have the dimension check moved into the graph run and the BasicBlocks be changed to use ArrayD?

liliatangxy (commented):
Left some initial comments, will look through the rest tonight

Edit: no additional comments

src/basic_block.rs
Vimalakirti: I'm also curious about @liliatangxy 's questions.

src/basic_block.rs
```diff
@@ -122,3 +141,94 @@ pub trait BasicBlock {
...
+    let model = arr_to_vec(model);
+    let model = model.iter().map(|x| x).collect();
+    let inputs = arr_to_vec(inputs);
+    let inputs = inputs.iter().map(|x| x).collect();
+    let outputs = self.basic_block.run(&model, &inputs);
```
Vimalakirti: Sometimes, we may want to run `self.basic_block.run(&model, &inputs);` N times inside a `BatchedBasicBlock` (as its name suggest(?)) in one ML layer. But currently, we have to create N `BatchedBasicBlock`s to do that. I am not sure if this is a good design choice.

src/basic_block/add.rs
```diff
@@ -5,6 +5,9 @@ use rand::rngs::StdRng;
 
 pub struct AddBasicBlock;
 impl BasicBlock for AddBasicBlock {
+  fn get_dims(&self) -> (Vec<usize>, Vec<usize>) {
+    (vec![], vec![1, 1])
```
Vimalakirti, line 9: We sometimes add model and input together instead of two inputs here. Shall we create an AddModelBasicBlock or  just generalize the current one? (same concerns for other basicblocks like Mul)

Vimalakirti (commented):
The main feature added is BatchedBasicBlock, which seems to be an ndarray wrapper for BasicBlock. The additions in other files are mostly fn get_dim() to indicate what the basicblock should take as its input data. 
The summary of this review is as follows:
- (same as Lilia's questions) Do we consider to support ArrayD in BasicBlock? If not, what are the concerns?
- Do we want to generalize BatchedBasicBlock to perform ML layer-level operations in one block?
- Do we want to generalize some BasicBlocks like Add and Mul to support more possible cases during model inference (e.g., model+input instead of input+input)?
Thanks

src/basic_block/add.rs
ddkang: Make an issue for this

ddkang: @Vimalakirti Did you make an issue for this?

ddkang: If so, resolve this comment

src/basic_block.rs
ddkang: Is this the layer issue? If so, raise an issue and resolve this @Vimalakirti

src/basic_block.rs
ddkang: @liliatangxy is this resolved with the layer thing?

src/basic_block.rs
ddkang: Is this resolved with the layer issue? @liliatangxy

src/basic_block/alternating.rs
ddkang: @liliatangxy is this resolved? If not, raise an issue

src/basic_block/transpose.rs
ddkang: @Vimalakirti please raise an issue or resolve this

src/graph.rs
arifiorino: For now we are using ArrayD in `BatchedBasicBlock` which will eventually be changed to a `Layer`

src/basic_block/concat.rs
arifiorino: The functions are not exactly the same, note that the one in `MatMul` is not the same as the one in `Alternating`.

src/basic_block/max.rs
arifiorino: For now this is unnecessary.

src/tests.rs
arifiorino: Right now all the basic blocks that have a prove/verify function are tested in this file.
The other blocks don't need to be tested because they don't require any proof/verification.

src/basic_block/alternating.rs
arifiorino: Yes, in the future we will repeat the same alpha and beta across blocks and precompute these MSMs.

src/graph.rs
arifiorino: No, this is not a good idea. This is because the model is potentially repeated across multiple nodes. For example there is one model for the cq relu table that only needs to be defined once.

src/main.rs
arifiorino: Yes, the helper function is called `gen_cq_table`. This is already defined.

src/basic_block/alternating.rs
arifiorino: For now this code is only in a few basic blocks so it does not need to be generalized.

Vimalakirti (approved):

src/basic_block.rs
liliatangxy: As discussed in the meeting I think we can still operate with Vec in BasicBlock but at the layer level we work with ArrayD

src/basic_block.rs
liliatangxy: Yes, resolved

PR #16, opened 2024-03-28
Fixed matmul

liliatangxy:
Added matmul with a fixed matrix. The only change necessary to cqlin is RLC for the input and output row commitments. Like the RLC in Aurora, it adds msms to the verifier.

cc single thread
log n | 1xn times nxn proof time (ms)
-- | --
3 | 7.14458
4 | 8.592204
5 | 11.358949
6 | 16.28443
7 | 22.655176
8 | 36.297327
9 | 60.246403
10 | 98.277973

Also moves `calc_pow` into utils


src/basic_block/matmul_fixed.rs
```diff
@@ -6,24 +6,36 @@ use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
...
 use ark_poly::{EvaluationDomain, GeneralEvaluationDomain, Polynomial};
-use ark_std::{UniformRand, Zero};
+use ark_std::{One, UniformRand, Zero};
 use ndarray::ArrayD;
 use rand::{rngs::StdRng, SeedableRng};
 use rayon::prelude::*;
 
-pub struct CQLinBasicBlock;
-impl BasicBlock for CQLinBasicBlock {
+fn calc_pow(alpha: Fr, n: usize) -> Vec<Fr> {
```
ddkang: How does the verifier know this was computed correctly, or is this done by both?

liliatangxy: The verifier obtains the shared rng object and should get the same alpha as the prover, and then computes the powers of alpha.

Vimalakirti (approved):

src/basic_block.rs
```diff
@@ -12,12 +12,12 @@ pub use concat::ConcatBasicBlock;
 pub use constant::ConstBasicBlock;
 pub use cq::CQBasicBlock;
 pub use cq2::CQ2BasicBlock;
-pub use cqlin::CQLinBasicBlock;
 pub use div::{DivConstBasicBlock, DivScalarBasicBlock};
 pub use eq::EqBasicBlock;
 pub use exp::ExpBasicBlock;
 pub use log::LogBasicBlock;
 pub use matmul::MatMulBasicBlock;
+pub use matmul_fixed::MatMulFixedBasicBlock;
```
arifiorino: I think we should keep the name of this as `CQLinBasicBlock` because that's what this basic block is based off of. It's cqlin with the combination idea from Aurora.

arifiorino:
We need to rebase this over the llama changes.

arifiorino (changes requested):

liliatangxy:
Finished the merge to master and rename, please review

src/basic_block/cqlin.rs
```diff
@@ -6,26 +6,39 @@ use ark_bn254::{Bn254, Fr, G1Affine, G1Projective, G2Affine, G2Projective};
 use ark_ec::pairing::Pairing;
 use ark_ff::Field;
 use ark_poly::{EvaluationDomain, GeneralEvaluationDomain, Polynomial};
-use ark_std::{UniformRand, Zero};
+use ark_std::{One, UniformRand, Zero};
+use ndarray::ArrayD;
 use rand::{rngs::StdRng, SeedableRng};
 use rayon::prelude::*;
 
+fn calc_pow(alpha: Fr, n: usize) -> Vec<Fr> {
```
ddkang: Is this reused elsewhere?

liliatangxy: Yeah, it is currently used in multiple files. Should I move it to utils in this pr?

ddkang: I think this PR is small enough that it's okay to do that now

arifiorino:
This looks good, except that `cargo run` fails now because it tries to pass a 1d vector times a 2d vector into cqlin. So this error needs to be fixed before this can be merged.

Vimalakirti (approved):
Except for Ari's request, I have no other concerns.

arifiorino:
This is the patch you need to apply to fix this:
[fix.patch](https://github.com/uiuc-kang-lab/zkml_proofs/files/14938291/fix.patch)

liliatangxy:
Added the patch

arifiorino:
Ok, you also moved `calc_pow` to the `util.rs` file. The issue is that `calc_pow` starts at `alpha^1` in `AlternatingBasicBlock` for a specific security reason. As it is now, the prover can swap the first two elements in `AlternatingBasicBlock`. In order to fix this, you need to change `calc_pow` to start at `alpha^1` in `util.rs`.

liliatangxy:
Makes sense, fixed

arifiorino (approved):
Ok, this looks good to me

PR #24, opened 2024-04-12
Basic Block NDArray + Simplification

arifiorino:
- BasicBlocks now run on `ArrayD` for their inputs and outputs.
  - There is no unnecessary copying of data since the MSMs run on slices of the `ArrayD`.
- `Exp`,`Log`,`ReLU`,`Sqrt`, and `DivConst` have all been defined simultaneously through a macro.
- `Alternate`, `Combine`, `Alternating`, `Concat`, and `Transpose` have been generalized into a `Permute` basic block that can do any of these operations.
- Assertions have been put in each basic block to check the shape of the model and inputs.
- `graph` and unit tests have been updated as well.


src/main.rs
```diff
@@ -3,45 +3,26 @@
...
-      } else {
-        x.map_axis(Axis(x.ndim() - 2), |y| Data::new(srs, &y.into_owned().into_raw_vec()))
-      }
-    })
-    .collect()
-}
-
 fn main() {
-  let srs = &ptau::load_file("challenge", 7);
+  let srs = &ptau::load_file("challenge14", 14);
```
Vimalakirti: There exists no `challenge14`. Therefore, `cargo run` fails, but replacing this line with `let srs = &ptau::load_file("challenge", 7);` does work.

src/tests.rs
```diff
@@ -1,95 +1,83 @@
...
+  let model = model.map(|x| DataEnc::new(srs, x));
+  let inputs: Vec<ArrayD<DataEnc>> = inputs.iter().map(|x| (*x).map(|y| DataEnc::new(srs, y))).collect();
+  let inputs: Vec<&ArrayD<DataEnc>> = inputs.iter().map(|input| input).collect();
+  let outputs: Vec<ArrayD<DataEnc>> = outputs.iter().map(|x| (*x).map(|y| DataEnc::new(srs, y))).collect();
+  let outputs: Vec<&ArrayD<DataEnc>> = outputs.iter().map(|output| output).collect();
   basic_block.verify(srs, &model, &inputs, &outputs, proof, &mut rng2);
 }
 
 #[test]
 fn testBasicBlocks() {
```
Vimalakirti, line 41: Not urgent for now, but we may consider to separate this test into small ones so that each test serves for just one basicblock.

src/basic_block/max.rs
Vimalakirti, line 1: How do we prove `max` currently?

Vimalakirti (commented):
I think this PR is exactly the one Lilia and I was looking for:
- Now all basicblocks support NDArray operations, which helps us to build Layer on top of them.
- Most nonlinear basicblocks are now simplified by a macro, which is more clear and also easier for us to maintain.

I summarize my 2 comments here, please check.
Except for them, I have no concern with other changes.
- how do we prove `max`?
- split `testBasicBlock` into multiple smaller tests since it is quite a big test now.

Besides, I will rebase my ongoing PR on this branch since it helps build `Layer`, which also involves consideration of tensor shape a lot.

src/main.rs
arifiorino: Oh yeah, sorry about that. I fixed it.

src/basic_block/max.rs
arifiorino: We don't need to prove `max` currently. This is because `max` is only used to implement the log sum exp trick. In this case `max` is subtracted and then added back in so it does not need to be proven, it could be any integer as far as the verifier is concerned.

src/tests.rs
arifiorino: Yes, potentially we can do this later.

src/basic_block/max.rs
Vimalakirti: I see, thanks! But we will need it eventually since ONNX operation does include `max`. Let us put this in mind.

src/tests.rs
Vimalakirti: I created a new [issue](https://github.com/uiuc-kang-lab/zkml_proofs/issues/25#issue-2241290327) for this and resolved this comment.

Vimalakirti (approved):

src/basic_block/cqlin.rs
```diff
@@ -7,32 +7,23 @@ use ark_ec::pairing::Pairing;
...
 impl BasicBlock for CQLinBasicBlock {
-  fn get_dims(&self) -> (Vec<usize>, Vec<usize>) {
-    (vec![2], vec![2])
+  fn run(&self, model: &ArrayD<Fr>, inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
+    assert!(model.ndim() == 2 && inputs.len() == 1 && inputs[0].ndim() == 2 && inputs[0].shape()[1] == model.shape()[0]);
+    let (a, b) = (
+      model.view().into_dimensionality::<Ix2>().unwrap(),
+      inputs[0].view().into_dimensionality::<Ix2>().unwrap(),
+    );
+    //vec![a.dot(&b).into_dyn()]
```
liliatangxy: remove comment

ddkang:
Add max as an issue @Vimalakirti . @liliatangxy can you take a quick look?

src/basic_block/permute.rs
```diff
@@ -0,0 +1,182 @@
...
+use ark_std::{ops::Mul, ops::Sub, One, UniformRand, Zero};
+use ndarray::{Array, ArrayD, Axis, IxDyn};
+use rand::{rngs::StdRng, SeedableRng};
+
+pub struct PermuteBasicBlock {
+  pub permutation: (Vec<usize>, Vec<usize>),
+}
+// Permute elements of a 2d matrix into another 2d matrix
+// LHS: inputs[0][i,j]  * alpha^(i+nj)
+// RHS: outputs[0][i,j] * alpha^(p[i]+p[j])
```
liliatangxy: Why do we not need both alpha and beta for the transpose case?

arifiorino: We don't need it because it's wrapped in to the powers of alpha.
This is the equation that's proven:
Screenshot 2024-04-13 at 10 51 26 AM
The permutation variable in the code corresponds to `(p,q)`.
In order to do a transpose you set `p=[0,n,2n,...]` and `q=[0,1,2,...]`

ddkang: It may be worth leaving a comment in the code as to why it's not needed

liliatangxy: Seems ok then. The comment after already shows this expression

liliatangxy (approved):

Vimalakirti:
> Add max as an issue @Vimalakirti . @liliatangxy can you take a quick look?

Added new issue here:
https://github.com/uiuc-kang-lab/zkml_proofs/issues/26#issue-2241589006

@arifiorino We think the PR is ready to be merged.

src/basic_block/permute.rs
arifiorino: I changed the comment to make it more clear as to why it's not needed.

PR #27, opened 2024-04-15
Add layer features

Vimalakirti:
Dear all, this PR is my proposal for realizing the concept of `Layer`. Please check `layer.rs`, `graph.rs` and files under `layer` to get the main ideas. I've made the following changes:
- Rename original `Graph` to `Subgraph`, which can be viewed as a "layer" in ML models. It supports `run`, `prove`, and `verify` as we did before. But note that it does not support `setup` since we do not want to setup the same basic block(s) across different layers. The structures of `Softmax` subgraph were contributed by @arifiorino (the python script), I rewrote them in Rust and made some changes to consider shape constraint in each basic block.
- Add `Graph` (i.e., overall ML model). It supports `setup` and high-level `run`, `prove`, and `verify`. Here "high-level" means every adjacency link between any two `Subgraphs` can be directly loaded from an ONNX file.
- Add `Layer` trait to realize
    - `load_layer_nodes`: create subgraph by layout the connection of `Node` (i.e., a basic block)
    - `consume_basic_block`: track the basicblocks the layer consumes
    - `layer_output_node`: return the output node id in each subgraph
 
Other minor changes:
- Add fn `name` under each basic block to track usage of basic blocks in the model
- Add basicblocks for `squeeze` and `unsqueeze`, currently I don't prove them as they are just helpers. (But we may need to, let's discuss later)
- I made an example in `main`, showing a subgraph can be created by calling `Layer` or directly implementing the layout as we did before.
- Add a new line between every `fn`
- `cargo run` and `cargo test` pass


src/graph.rs
```diff
@@ -10,57 +10,112 @@ pub struct Node {
 
 pub struct Graph {
   pub basic_blocks: Vec<Box<dyn BasicBlock>>,
-  pub nodes: Vec<Node>,
+  pub layers: Vec<Box<dyn Layer>>,
+  pub layer_nodes: Vec<Vec<Node>>,          // each Vec<Node> is a layer
```
liliatangxy: Is there a reason why this is part of the graph and not the layers?

Vimalakirti: Thanks, I've updated this struct.

Vimalakirti:
New changes:

- Simplification: remove `Subgraph` and put the`run`, `prove`, `verify` functions into `Layer` trait.
- Encapsulation: `Graph` (model) --> `Layer` (ONNX layer) --> `Node` (zk basicblocks).
- Prevent repeated setup: Add `Hash` for dyn BasicBlock to use `HashMap<&Box<dyn BasicBlock>, usize>` while creating layer nodes

src/basic_block.rs
```diff
@@ -110,3 +126,18 @@ pub trait BasicBlock {
     ()
   }
 }
+
+// hash for trait basicblock
+impl std::hash::Hash for dyn BasicBlock {
+  fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+    self.name().hash(state);
```
liliatangxy, line 137: Do we have to distinguish basic blocks with different models (e.g. table for CQ) in the hash fn?

Vimalakirti: Thanks!
Yes, we need. But it's already done (attached fig):
image
It means we can put "ReLU"/"Log"/... as the name of different CQ blocks; and the hash will be different accordingly.

liliatangxy: I see, I think it was missing from CQ but CQ2 and CQLin addresses this.

Vimalakirti: Thanks, I forgot to address this for CQ as we don't use it currently. I've fixed it.

ddkang: If this is confusing, we should document it in the code

Vimalakirti: Thanks, I've added comments to explain this in the code

ddkang: This is a bad comment. You should explain _why_ the code is doing what it's doing, not some canned comment that can be inferred from the trait implementation.

Vimalakirti: Thanks, no problem, I've edited the comments

Vimalakirti:
I closed this PR because there already exists another layer [PR](https://github.com/uiuc-kang-lab/zkml_proofs/pull/31).

PR #28, opened 2024-04-21
Removing extraneous recomputation

arifiorino:
- Added `encodeOutputs` method to `BasicBlock`
  - Defaults to running an fft and an msm
- Implemented `encodeOutputs` for `AddBasicBlock` which adds the `g1` commitments of the inputs
  - This avoids recomputing the fft and msm for the output
- Updated main and unit tests

### DIFF
--- a/src/basic_block.rs
+++ b/src/basic_block.rs
@@ -85,12 +85,20 @@ pub trait BasicBlock {
   fn run(&self, _model: &ArrayD<Fr>, _inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
     vec![]
   }
+
+  // This function encodes the outputs of the BasicBlock into Data objects.
+  // It defaults to running Data::new() on the last dimension of the outputs which runs an FFT and an MSM.
+  // But for certain basic blocks such as add and reshape, this can be done much faster, and it should be overriden in these cases.
   fn encodeOutputs(&self, srs: &SRS, _model: &ArrayD<Data>, _inputs: &Vec<&ArrayD<Data>>, outputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Data>> {
     outputs.iter().map(|x| util::convert_to_data(srs, x)).collect()
   }
+
+  // The subsequent setup/prove/verify functions run on encoded Data objects (vector commitments).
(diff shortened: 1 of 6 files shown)


src/basic_block.rs
```diff
@@ -84,6 +85,9 @@ pub trait BasicBlock {
   fn run(&self, _model: &ArrayD<Fr>, _inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
     vec![]
   }
+  fn encodeOutputs(&self, srs: &SRS, _model: &ArrayD<Data>, _inputs: &Vec<&ArrayD<Data>>, outputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Data>> {
```
ddkang, line 92: new lines

ddkang: Why do we need a default implementation?

ddkang:
Can you explain what encode outputs does?

src/basic_block.rs
arifiorino: I added the new lines.

arifiorino: We need a default implementation because for most basic blocks, the fastest way is to just encode the outputs into `Data` objects via `Data::new()`. But there are certain basic blocks where it can be done much faster such as add and reshape.

arifiorino:
I sent a message on slack with the explanation and I added a comment to the code. I can also rewrite the comment if it's not written clearly.

src/basic_block.rs
ddkang: This is a good start but you should say why we encode - something about caching/reducing computation because...

arifiorino: I added a comment explaining that.

Vimalakirti (approved):

liliatangxy (approved):
I think `SubBasicBlock` can also benefit?

ddkang:
@arifiorino feel free to merge but raise an issue or address Lilia's comments

arifiorino:
I implemented subtract as well. @liliatangxy are we good to merge?

liliatangxy:
Yeah looks good

ddkang (approved):

PR #29, opened 2024-04-24
Batch pairings

### LINKED ISSUE #22: Batching pairings
We can look into pairing batching already done by arkworks: https://docs.rs/ark-ec/0.4.0/ark_ec/pairing/trait.Pairing.html#method.multi_pairing

@liliatangxy will look into what plonk/halo2 is doing

arifiorino:
Combines pairings that have the same G1 or G2 element for faster verification.
- `verify` now returns `Vec<PairingCheck>` with `PairingCheck = Vec<(G1Affine, G2Affine)>`. All the pairings in a `PairingCheck` should sum to zero.
- `util::combine_pairing_checks` combines multiple pairing checks using a greedy algorithm in `O(nlogn)`.
- All basic blocks have been modified to return their pairing checks.
- `graph.rs` and unit tests have been updated accordingly.

Closes #22


src/basic_block.rs
```diff
@@ -119,7 +119,7 @@ pub trait BasicBlock {
     _outputs: &Vec<&ArrayD<DataEnc>>,
     _proof: (&Vec<G1Affine>, &Vec<G2Affine>),
     _rng: &mut StdRng,
-  ) {
-    ()
+  ) -> Vec<Vec<(G1Affine, G2Affine)>> {
```
ddkang: Make this a typedef

src/util.rs
```diff
@@ -138,3 +144,74 @@ pub fn convert_to_data(srs: &SRS, a: &ArrayD<Fr>) -> ArrayD<Data> {
   }
   a.map_axis(Axis(a.ndim() - 1), |r| Data::new(srs, r.as_slice().unwrap()))
 }
+
+pub fn combine_pairing_checks(checks: &Vec<&Vec<(G1Affine, G2Affine)>>) -> (Vec<G1Affine>, Vec<G2Affine>) {
```
ddkang: Does this do the actual verification? If so, should be made clear in the function name

arifiorino: No, it doesn't do the verification, it just combines the pairing checks. The verification is done in `graph.rs`.

ddkang:
Can we get a benchmark on the first layer of Llama2?

src/basic_block.rs
arifiorino: I will do that. Just to be clear, we said that `Vec<(G1Affine, G2Affine)>` should be made into `PairingCheck`. And therefore this function will return `Vec<PairingCheck>` right?

arifiorino:
Yes, I'm working on it.

src/basic_block.rs
ddkang: I think that works

arifiorino: Ok, I changed it to a typedef.

Vimalakirti (approved):
This PR introduces `util::combine_pairing_checks` to combine group elements for pairing checks across separate proofs into a single `ParingCheck`, which reduce the number of pairing required during verification phase.
-
I've reviewed and have no concern for the PR but just came up with a random thought:
Currently we collect all the G1 and G2 elements and then combine them through `combine_pairing_checks`. What if we combine them every time when new G1 and G2 elements are pushed in (i.e., one call of basicblock.verify)? Will it make things easier/faster?

ddkang:
Looks good to merge, raise an issue to consider it

liliatangxy (approved):

arifiorino:
@Vimalakirti That does not make a difference. It will be the same amount of time either way.

Vimalakirti:
@arifiorino   Thanks! I thought we may have no need to maintain BinaryTree in that case since S.len() would always be 1. But maybe it's wrong and it's not worth considering it for now. Feel free to merge this PR. (I still created an [issue](https://github.com/uiuc-kang-lab/zkml_proofs/issues/30); we can delete it later if it's wrong)

PR #31, opened 2024-04-29
Layer with reusable weights and tables

liliatangxy:
Enables reuse of weights and table on top of the layer PR #27 and avoids recomputation for their setups. This also simplifies the handling of weights and tables for the user into providing a weights `HashMap` and configuration of tables by layer. 

Changes:
- Weights are stored in a `HashMap`, and are identified with a user-provided key provided to the `LayerConfig`
- `Graph` has a constructor that handles table creation based on a list of `LayerConfigs`
- Added `Setup` object and moved setup logic. Setups and weights/table `Data` objects are constructed once for each unique hash map entry
- Added `BasicBlockType`, `LayerType`, and for basic blocks `block_type()`, `weights_name()`
- Removed `setup` from `layer` and lower, changed signatures for `run`, etc., to not use `model` when possible


ddkang:
Raise issues for the last 3 notes

src/basic_block.rs
```diff
@@ -89,24 +90,53 @@ impl DataEnc {
...
+  }
+
   fn name(&self) -> String {
     "".to_string()
   }
 
-  fn run(&self, _model: &ArrayD<Fr>, _inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
-    vec![]
+  fn weights_name(&self) -> String {
+    "".to_string()
```
ddkang: Why don't we just error here?

src/graph.rs
```diff
@@ -9,14 +11,69 @@ pub struct Node {
...
+
+    let mut basic_blocks = vec![];
+    for (i, layer) in layers.iter().enumerate() {
+      basic_blocks.append(&mut layer.consume_basic_block(&configs[i]));
+    }
+
+    let mut table_map: HashMap<String, ArrayD<Fr>> = HashMap::new();
+    for i in 1..basic_blocks.len() {
+      let table = match (basic_blocks[i - 1].block_type(), basic_blocks[i].block_type()) {
+        (BasicBlockType::ChangeSF, BasicBlockType::CQ2)
```
ddkang: What's going on here?

src/graph.rs
```diff
@@ -9,14 +11,69 @@ pub struct Node {
...
+      layers,
+      layer_configs: configs,
+      layer_inputs,
+      basic_blocks,
+      weights_map: weights,
+      table_map,
+    }
+  }
+
+  pub fn run(&self, inputs: &Vec<&ArrayD<Fr>>) -> Vec<Vec<Vec<ArrayD<Fr>>>> {
```
ddkang, line 106: What's going on with the vec vec vec stuff?

src/setup.rs
```diff
@@ -0,0 +1,181 @@
+use crate::util::{self, convert_to_data};
```
ddkang: Why is this a generic setup file?

ddkang:
Can we make a GitHub action to run tests when commits are pushed? Can raise an issue

src/graph.rs
Vimalakirti: We need to initialize a CQ2 table for every nonlinear basicblock. With this realization, it's great that we no longer mix model weights and lookup tables together. But we could refactor codes (e.g., pack nonlinear basikblocktypes somewhere else) or add comments to make it more clear.

Vimalakirti (approved):
Thanks! Except for readability, GitHub testing, and CustomLayer (we can delete it if unnecessary) issues, I have no other concerns. 
Btw, shall I close my previous PR since this PR is directly based on that one?

src/graph.rs
liliatangxy: The indices are [layer no.][node id in that layer][the output index of that node].

@Vimalakirti Is there a reason why we need the output index?

Vimalakirti: @liliatangxy we need that because some nodes may return two outputs (e.g., Split). But currently I think we don't have this kind of basicblock.

liliatangxy:
Cleaned up the graph constructor, added error checking for weights name and CQ2 following a nonlinearity, and moved setup back.
`name()` is actually used for the `Hash` trait, which is needed in the codebase so we don't need to remove it.

I think supporting `CustomLayer` would require us to put `Layer` inside the `LayerConfig`.

src/graph.rs
liliatangxy: Ok, I think we can keep that in the code. I can add a comment to clarify this

src/graph.rs
liliatangxy: I simplified the logic here by checking whether a BB is a nonlinearity and erroring if it is not followed by CQ2. Right now table generation only uses CQ2 (combining range check with the nonlinearity), can introduce CQ logic in another PR.

Vimalakirti:
Hi, I found two bugs in `softmax` and `graph`:
- basicblock should not duplicate in graph initialization
- the output node of softmax is incorrect

I've pushed my patch for them, please check.

liliatangxy:
Thanks for catching them, the changes look good

src/basic_block/cq2.rs
```diff
@@ -15,20 +19,36 @@ use rand::{rngs::StdRng, SeedableRng};
...
-  fn setup(&self, srs: &SRS, model: &ArrayD<Data>) -> (Vec<G1Projective>, Vec<G2Projective>) {
-    assert!(model.ndim() == 1 && model.len() == 2);
-    let N = model[0].raw.len();
+  fn block_type(&self) -> BasicBlockType {
+    BasicBlockType::CQ2
+  }
+
+  fn name(&self) -> String {
+    // concat "CQ2" with self.name
+    format!("CQ2({})", self.name)
```
arifiorino, line 34: This is unnecessary because the `Debug` trait already includes `CQ2` and `self.name`.

src/basic_block.rs
```diff
@@ -80,24 +89,66 @@ impl DataEnc {
...
+  Sqrt,
+  Squeeze,
+  Sub,
+  Sum,
+  Unsqueeze,
+}
+
+pub trait BasicBlock: std::fmt::Debug {
+  fn block_type(&self) -> BasicBlockType {
+    BasicBlockType::Constant
```
arifiorino: Why would the default be `Constant`? I think there should be no default and every basic block should define this. But there is probably a more efficient way of doing this as well (maybe using `format!()` so we don't have to add code to every basic block.

src/basic_block.rs
```diff
@@ -80,24 +89,66 @@ impl DataEnc {
...
   }
-  fn setup(&self, _srs: &SRS, _model: &ArrayD<Data>) -> (Vec<G1Projective>, Vec<G2Projective>) {
-    (Vec::new(), Vec::new())
+
+  fn name(&self) -> String {
```
arifiorino, line 127: We should think about if we even need this method, because it might be better to just use `format!()` instead everywhere this method is used.

src/basic_block.rs
```diff
@@ -80,24 +89,66 @@ impl DataEnc {
...
+    format!("{self:?}")
+  }
+
+  fn weights_name(&self) -> Result<String, String> {
```
arifiorino, line 131: I thought the weights would be part of the `name` itself.

src/basic_block.rs
```diff
@@ -110,3 +161,22 @@ pub trait BasicBlock {
...
+/// the usage of BasicBlocks in a graph. This prevents redundant setups for
+/// the same BasicBlock (e.g., we use CQ2 for ReLU and Log, but we don't
+/// want to setup CQ2 twice for ReLU)
+impl std::hash::Hash for dyn BasicBlock {
+  fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+    self.name().hash(state);
+  }
+}
+
+impl std::cmp::PartialEq for dyn BasicBlock {
```
arifiorino, line 188: This could also be done by adding `#[derive(Debug, PartialEq)]` to the basic blocks which might be more intuitive.

src/graph.rs
```diff
@@ -1,66 +1,256 @@
...
+  pub layers: Vec<Box<dyn Layer>>,
+  pub layer_configs: Vec<LayerConfig>,
+  pub layer_inputs: Vec<Vec<(i32, usize)>>, // ONNX graph (layer id, layer slot)
   pub basic_blocks: Vec<Box<dyn BasicBlock>>,
-  pub nodes: Vec<Node>,
+  pub weights_map: HashMap<String, ArrayD<Fr>>,
+  pub table_map: HashMap<String, ArrayD<Fr>>,
 }
 
 impl Graph {
```
arifiorino, line 55: Shouldn't we just convert the layers into a `Graph`, and leave the `Graph` struct unchanged?

src/layer.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+#[derive(Debug)]
+pub struct LayerConfig {
+  pub layer_type: LayerType,
+  pub input_params: HashMap<String, usize>,
+  pub weights_names: Vec<String>,
+}
+
+pub trait Layer {
```
arifiorino, line 35: So we have `run/prove/verify` defined for `BasicBlock`,`Layer`, and `Graph`? we don't need it defined for all these three.

src/layer/add.rs
```diff
@@ -0,0 +1,20 @@
+use super::{Layer, LayerConfig};
```
arifiorino, line 1: This is too many lines (20) that we will have to write for every layer (there are 154 layers in the tract library)

src/graph.rs
```diff
@@ -1,66 +1,256 @@
...
+
+pub struct CQSetup {
+  pub table: ArrayD<Data>,
+  pub Q_i_x_1: Vec<Vec<G1Affine>>,
+  pub L_i_x_1: Vec<G1Affine>,
+  pub L_i_0_x_1: Vec<G1Affine>,
+  pub T_x_2: Vec<G2Affine>,
+}
+
+pub enum SetupType {
```
arifiorino, line 36: This seems a bit hardcoded to me. There might be a more flexible/simpler way to do this.

src/basic_block/cq.rs
```diff
@@ -15,18 +18,34 @@ use rand::{rngs::StdRng, SeedableRng};
...
 impl BasicBlock for CQBasicBlock {
-  fn setup(&self, srs: &SRS, model: &ArrayD<Data>) -> (Vec<G1Projective>, Vec<G2Projective>) {
-    assert!(model.len() == 1);
-    let model = model.first().unwrap();
-    let N = model.raw.len();
+  fn block_type(&self) -> BasicBlockType {
+    BasicBlockType::CQ
+  }
+
+  fn name(&self) -> String {
```
arifiorino, line 32: This is unnecessary because the `Debug` trait already includes `CQ` and `self.name`.

src/basic_block/cqlin.rs
```diff
@@ -37,6 +55,9 @@ impl BasicBlock for CQLinBasicBlock {
     util::ifft_in_place(domain_m, &mut L_V_i_x_n);
     let mut L_V_i_x: Vec<G1Projective> = srs.X1P[..m].into_par_iter().map(|x| (*x).into()).collect();
     util::ifft_in_place::<G1Projective>(domain_m, &mut L_V_i_x);
+    let L_H_i_x = L_H_i_x.iter().map(|x| (*x).into()).collect();
+    let L_V_i_x_n = L_V_i_x_n.iter().map(|x| (*x).into()).collect();
+    let L_V_i_x = L_V_i_x.iter().map(|x| (*x).into()).collect();
```
arifiorino, line 58: Earlier, the setup functions returned `(Vec<G1Projective>,Vec<G2Projective>)` so we didn't have to write conversions like this into the setup functions.

src/basic_block/cq2.rs
```diff
@@ -37,28 +57,41 @@ impl BasicBlock for CQ2BasicBlock {
...
     let mut L_i_x_1 = srs.X1P[..N].to_vec();
     util::ifft_in_place(domain_N, &mut L_i_x_1);
     let mut L_i_0_x_1 = L_i_x_1.clone();
     let temp = srs.X1P[N - 1] * Fr::from(N as u64).inverse().unwrap();
     L_i_0_x_1.par_iter_mut().enumerate().for_each(|(i, x)| *x = *x * domain_N.group_gen_inv().pow(&[i as u64]) - temp);
-    setup.extend(L_i_x_1);
-    setup.extend(L_i_0_x_1);
-    return (setup, setup2);
+    let L_i_x_1 = L_i_x_1.iter().map(|x| (*x).into()).collect();
+    let L_i_0_x_1 = L_i_0_x_1.iter().map(|x| (*x).into()).collect();
```
arifiorino, line 67: Earlier, the setup functions returned `(Vec<G1Projective>,Vec<G2Projective>)` so we didn't have to write conversions like this into the setup functions.

src/basic_block/cq.rs
```diff
@@ -35,40 +54,52 @@ impl BasicBlock for CQBasicBlock {
...
     util::ifft_in_place(domain_N, &mut L_i_x_1);
     let mut L_i_0_x_1 = L_i_x_1.clone();
     let temp = srs.X1P[N - 1] * Fr::from(N as u64).inverse().unwrap();
     L_i_0_x_1.par_iter_mut().enumerate().for_each(|(i, x)| *x = *x * domain_N.group_gen_inv().pow(&[i as u64]) - temp);
-    let mut setup = Q_i_x_1;
-    setup.extend(L_i_x_1);
-    setup.extend(L_i_0_x_1);
-    return (setup, vec![T_x_2]);
+    let L_i_x_1 = L_i_x_1.iter().map(|x| (*x).into()).collect();
+    let L_i_0_x_1 = L_i_0_x_1.iter().map(|x| (*x).into()).collect();
```
arifiorino, line 64: Earlier, the setup functions returned `(Vec<G1Projective>,Vec<G2Projective>)` so we didn't have to write conversions like this into the setup functions.

src/basic_block.rs
```diff
@@ -80,24 +89,66 @@ impl DataEnc {
...
+
+  fn weights_name(&self) -> Result<String, String> {
+    Err(format!("{:?} does not use weights.", self.block_type()))
   }
+
+  fn setup(&self, _srs: &SRS, _model: &ArrayD<Fr>) -> SetupType {
+    SetupType::None
+  }
+
+  fn run(&self, _weights: &ArrayD<Fr>, _inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
```
arifiorino, line 139: Why change this to `weights`? It could either be the weights, or it could be tables.

src/basic_block.rs
```diff
@@ -80,24 +89,66 @@ impl DataEnc {
...
+
+  fn name(&self) -> String {
+    format!("{self:?}")
+  }
+
+  fn weights_name(&self) -> Result<String, String> {
+    Err(format!("{:?} does not use weights.", self.block_type()))
   }
+
+  fn setup(&self, _srs: &SRS, _model: &ArrayD<Fr>) -> SetupType {
```
arifiorino, line 135: I thought we weren't using `model` anymore.

src/basic_block.rs
```diff
@@ -110,3 +161,22 @@ pub trait BasicBlock {
     ()
   }
 }
+
+/// Hash, PartialEq, Eq are implemented for dyn BasicBlock trait objects
+/// so that BasicBlock can be used in HashMaps, which is useful for tracking
+/// the usage of BasicBlocks in a graph. This prevents redundant setups for
+/// the same BasicBlock (e.g., we use CQ2 for ReLU and Log, but we don't
+/// want to setup CQ2 twice for ReLU)
+impl std::hash::Hash for dyn BasicBlock {
```
arifiorino, line 182: We could just define a `HashMap` with strings as keys instead, and run the `Debug` trait to generate the keys.

arifiorino (changes requested):
I left several detailed comments about this PR. In addition to these, this PR is not able to load an onnx file at all. There are several subtleties in loading an onnx file, and none of these are addressed here.

src/basic_block.rs
liliatangxy: For a basic block without weights, it will be an empty array

src/basic_block/cq2.rs
liliatangxy: `format` isn't sufficient for the basic blocks containing `HashTable` e.g., `CQ2`

arifiorino: The `table_dict` parameter in `CQ2BasicBlock` doesn't contain anything when we're setting it up. It only gets populated during proof time. Therefore it shouldn't be an issue.

PR #33, opened 2024-05-01
add broadcast support for AddLayer

Vimalakirti:
Many operations in ONNX require the broadcast feature such as Max, Add, Sub, Mul, And, and more.
This **draft PR** is to demo how to do broadcast `Add` in `AddLayer`. But the current PR is inefficient in that:

- many repeated codes can be moved to a helper function
- we call `to_owned()` for broadcasted `ArrayD`, which means new memory is required. I'm not sure how to do it more efficiently for now.
- (and may contain bugs)

I'll fix them later

### DIFF
--- a/src/layer/add.rs
+++ b/src/layer/add.rs
@@ -1,6 +1,11 @@
 use super::{Layer, LayerConfig};
 use crate::basic_block::BasicBlock;
-use crate::graph::Node;
+use crate::graph::{Node, Setup, SetupType};
+use crate::util::convert_to_data;
+use crate::DataEnc;
+use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
+use ndarray::{ArrayD, Axis, IxDyn, Zip};
+use rand::rngs::StdRng;
 use std::collections::HashMap;
 
 pub struct AddLayer;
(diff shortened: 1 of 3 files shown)


liliatangxy:
What are the cases of broadcasting we want to support?

Vimalakirti:
> What are the cases of broadcasting we want to support?

Thanks, this is a good question. Excuse me for not providing the resource.
Please check [this ONNX document](https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md)

src/layer/add.rs
```diff
@@ -25,4 +30,120 @@ impl Layer for AddLayer {
...
+    }
+    let broadcast_shape = a.raw_dim();
+    let ndim = a.ndim();
+    if ndim > 0 {
+      let a_lanes = a.lanes(Axis(ndim - 1));
+      let b_lanes = b.lanes(Axis(ndim - 1));
+      let c_lanes = outputs[0][0].lanes(Axis(ndim - 1));
+      
+      a_lanes.into_iter().zip(b_lanes.into_iter()).zip(c_lanes.into_iter()).for_each(|((a, b), c)| {
+      let a = a.to_owned().into_dyn();
```
liliatangxy, line 91: Would it make sense to change the `basic_block` `prove` and `verify` into using `ArrayViewD` instead of `ArrayD` to avoid copying?

Vimalakirti: It sounds like a good and simple idea. In this case, we also need to change inputs of `run` into `ArrayViewD`.

PR #35, opened 2024-05-02
Read ONNX files

arifiorino:
`run/setup/prove/verify` ONNX files

Changes:
- Load an onnx file via `proto_model_for_path()` and load the weights via `load_tensor()`
- Convert each onnx operation into a `Graph` object and combine the `Graph` objects via hashing
- Add a small two layer feed forward network `sample.onnx` which is proved in `main.rs`
- Add the `genModel` method to a basic block which allows a basic block to generate it's own model from it's parameters
- Changed all `arr0` tensors to `arr1` tensors with one element. This simplifies the code because we don't have to run `*a.first().unwrap()` to get `a`'s element and we can do `a[0]` instead. All assert statements have been updated accordingly.
- Changed `MatMulBasicBlock` to support matrix-vector multiplication
- Added `Layer`s


Vimalakirti:
Thanks for pushing the draft PR!
We had tried a similar way to realize ONNX converting before. For example, I think `onnx.rs` is similar with my `onnx_converter.rs` [here](https://github.com/uiuc-kang-lab/zkml_proofs/blob/905eeae9bb5c582eaad37c4c682ed452a2ede8ae/src/onnx_converter.rs). Please check [the closed PR](https://github.com/uiuc-kang-lab/zkml_proofs/pull/17) for more information. 

But as Daniel mentioned on Tue, I did encounter some similar issues before. For instance:

-  it would be really difficult for us to know which ONNX node is not converted properly when some `BasicBlock` crashes during proving/verification. For instance, when `SigmoidBasicBlock` crashes, we don't know which ONNX node (it can be the 1st or 3rd Sigmoid) or which type of layer (it can be sigmoid or softmax) we should check. (we may try indexing `Node` in this PR)
- while implementing in this way can save lots of codes, it will be overwhelming to maintain the extremely large function `load_file` eventually. (we may try to decouple them into smaller functions for every Op, but it might become something similar to the Layer PR)
- if some ML folks want to build an ML model using our repo, they always need to get an ONNX file first rather than build a simple one in `zkml_proofs` as we don't have layer. (without layer concept, I'm not sure how to make this repo less distant for them.)

arifiorino:
@Vimalakirti A layer object is unnecessary because it is identical in functionality to a `Graph`. A simple `HashMap` can point from a node in the `Graph` to a node in the onnx file for debugging purposes.

ddkang:
> @Vimalakirti A layer object is unnecessary because it is identical in functionality to a `Graph`. A simple `HashMap` can point from a node in the `Graph` to a node in the onnx file for debugging purposes.

https://github.com/uiuc-kang-lab/zkml_proofs/pull/35/files#diff-548b2d389ea17641c304001bef4e3d64f7470ec5644a51ae43cc495e3241ba34 this is an unacceptably bad way to program layers. There's no separation of concerns and it's extremely difficult to understand

src/basic_block.rs
```diff
@@ -82,7 +82,11 @@ impl DataEnc {
     };
   }
 }
-pub trait BasicBlock {
+pub trait BasicBlock: std::fmt::Debug {
```
ddkang, line 91: space

src/basic_block/cq.rs
```diff
@@ -9,18 +9,23 @@ use ark_std::{
...
 
+#[derive(Debug)]
 pub struct CQBasicBlock {
   pub table_dict: HashMap<Fr, usize>,
+  pub setup: Option<(i32, usize)>,
 }
 impl BasicBlock for CQBasicBlock {
+  fn genModel(&self) -> ArrayD<Fr> {
+    Array1::from_iter(self.setup.unwrap().0..self.setup.unwrap().0 + (self.setup.unwrap().1 as i32)).map(|x| Fr::from(*x)).into_dyn()
+  }
```
ddkang, line 26: new line

src/basic_block/cq2.rs
```diff
@@ -14,10 +14,19 @@ use rand::{rngs::StdRng, SeedableRng};
...
+  pub setup: Option<(Box<dyn BasicBlock>, i32, usize)>,
 }
 impl BasicBlock for CQ2BasicBlock {
+  fn genModel(&self) -> ArrayD<Fr> {
+    util::gen_cq_table(
+      &(self.setup.as_ref().unwrap().0),
+      self.setup.as_ref().unwrap().1,
+      self.setup.as_ref().unwrap().2,
+    )
+  }
```
ddkang, line 30: new line

src/onnx.rs
```diff
@@ -0,0 +1,90 @@
...
+    let name = tensor.name.clone();
+    let tensor = load_tensor(&*onnx.provider, &tensor, None).unwrap();
+    let tensor = tensor.into_array::<f32>().unwrap();
+    let tensor = tensor.map(|x| Fr::from((*x * ((1 << 3) as f32)).round() as i32));
+    outputs_idx.insert(name, vec![(graph.basic_blocks.len() as i32, 0)]);
+    graph.nodes.push(Node {
+      basic_block: graph.basic_blocks.len(),
+      inputs: vec![],
+    });
+    graph.basic_blocks.push(Box::new(ConstBasicBlock {}));
```
liliatangxy, line 36: I'm not sure if CQLinBasicBlock works with this, might need to change it to use input instead of `model`

arifiorino: The reason we can't change cqlin to use `inputs` instead of `model` is because the cqlin basic block needs to be set up based on the `model`, and `inputs` are assumed to be only known at proof time.

arifiorino: In the future we can prune the graph so that whenever there is a constant followed by a matmul, it is replaced by cqlin.

liliatangxy: Oh I meant in the basic block file itself, in `run`, `ConstBasicBlock` would pass in `model` as an `input` into `CQLinBasicBlock`, but `CQLinBasicBlock` `run` uses the `model`

liliatangxy: Sure I guess

arifiorino: Yeah but the assumption is that `run` only happens at proof time, not at setup time. I added a comment in the code addressing how we can fix this in the future.

liliatangxy (approved):

PR #36, opened 2024-05-14
Fiat-Shamir + Serialization

### LINKED ISSUE #13: Fiat-Shamir
Currently, the prover and verifier have a shared `Rng` that they use to generate the random values in the proofs. This `Rng` should be seeded with a cryptographic hash of the public values, in order to implement the Fiat-Shamir heuristic.

arifiorino:
- Saves `models`, `inputs`, `outputs`, and `proofs` to disk using serde and ark_serialize.
- The bytes from `models`, `inputs`, and `outputs` are hashed via Keccak-256, and then passed as the seed to the intial `rng`.
  - After each basic block is proven, the proof is hashed and passed into the `rng` as well.
- This makes the proof secure and non-interactive using the Fiat-Shamir heuristic.

Closes #13.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,19 +21,19 @@ mod tests;
 mod util;
 
 fn prove(srs: &SRS, graph: &mut Graph, models: Vec<ArrayD<Fr>>) {
-  //Run:
+  // Run:
   let input: Vec<_> = (0..4).into_par_iter().map_init(rand::thread_rng, |rng, _| Fr::from(rng.gen_range(-4..4))).collect();
   let input = arr1(&input).into_dyn();
   let inputs = vec![&input];
   let models = models.iter().map(|x| x).collect();
   let outputs = graph.run(&inputs, &models);
 
-  //Setup:
+  // Setup:
   let models: Vec<ArrayD<Data>> = models.iter().map(|model| convert_to_data(srs, model)).collect();
   let models: Vec<&ArrayD<Data>> = models.iter().map(|model| model).collect();
   let setups = graph.setup(srs, &models);
 
-  //Encode Data:
+  // Encode Data:
   let setups: Vec<(Vec<G1Affine>, Vec<G2Affine>)> =
     setups.iter().map(|x| (x.0.iter().map(|y| (*y).into()).collect(), x.1.iter().map(|y| (*y).into()).collect())).collect();
   let setups = setups.iter().map(|x| (&x.0, &x.1)).collect();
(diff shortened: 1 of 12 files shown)


liliatangxy (approved):

src/util.rs
```diff
@@ -104,6 +106,35 @@ pub fn toeplitz_mul<G: ScalarMul + std::ops::MulAssign<Fr>>(domain: GeneralEvalu
   r
 }
 
+pub fn ark_se<S, A: CanonicalSerialize>(a: &A, s: S) -> Result<S::Ok, S::Error>
```
ddkang, line 112: Is this a required name?

src/util.rs
```diff
@@ -104,6 +106,35 @@ pub fn toeplitz_mul<G: ScalarMul + std::ops::MulAssign<Fr>>(domain: GeneralEvalu
...
+where
+  S: serde::Serializer,
+{
+  let mut bytes = vec![];
+  a.serialize_compressed(&mut bytes).map_err(serde::ser::Error::custom)?;
+  s.serialize_bytes(&bytes)
+}
+
+pub fn ark_de<'de, D, A: CanonicalDeserialize>(data: D) -> Result<A, D::Error>
```
ddkang, line 121: Ditto about the name

src/main.rs
```diff
@@ -15,52 +20,91 @@ mod ptau;
...
-  let outputs: Vec<Vec<&ArrayD<Fr>>> = outputs.iter().map(|output| output.iter().map(|x| x).collect()).collect();
-  let outputs: Vec<&Vec<&ArrayD<Fr>>> = outputs.iter().map(|output| output).collect();
 
   //Setup:
   let models: Vec<ArrayD<Data>> = models.iter().map(|model| convert_to_data(srs, model)).collect();
   let models: Vec<&ArrayD<Data>> = models.iter().map(|model| model).collect();
   let setups = graph.setup(srs, &models);
-  //Converting to affine
+
+  //Encode Data:
```
ddkang: Spaces after comments everywhere

ddkang: Please follow standard programming conventions

src/util.rs
arifiorino: I can change the name to be anything, maybe `ark_serialize` would be better. The reason I have this function is because `ndarray` uses `serde` for serialization while `G1Affine` use `ark_serialize` for serialization. So I need a way to bridge between the two, and I'm using this code snippet: https://github.com/arkworks-rs/algebra/issues/178#issuecomment-1413219278.

arifiorino: Also maybe I can add a comment to the code explaining this and referencing a link to the code snippet.

ddkang: Both would be good

src/main.rs
arifiorino: Ok I fixed this

src/util.rs
arifiorino: Ok I added those

ddkang (approved):

PR #38, opened 2024-05-16
Cache alphas and betas

### LINKED ISSUE #23: Use same alpha/beta across aurora proofs
Use the same alpha/beta across proofs to reduce recomputation (msms) in prover and verifier

arifiorino:
- Added a `ProveVerifyCache` object which allows for caching across proofs and across verifications.
- Moved the `table_dict` object in `CQBasicBlock`/`CQ2BasicBlock` to the `ProveVerifyCache`.
- Combined the alphas and betas in `MatMul`/`CQLin`/`Permute` to a single alpha/beta, and used caching to prevent the prover/verifier from recomputing MSMs.
- There was an error when adding / subtracting vectors of all zero, which as been fixed.

Closes #23.

### DIFF
--- a/src/basic_block.rs
+++ b/src/basic_block.rs
@@ -2,8 +2,7 @@
 use crate::util::{self, ark_de, ark_se};
 pub use add::AddBasicBlock;
 use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
-use ark_poly::univariate::DensePolynomial;
-use ark_poly::{EvaluationDomain, GeneralEvaluationDomain};
+use ark_poly::{univariate::DensePolynomial, DenseUVPolynomial, EvaluationDomain, GeneralEvaluationDomain};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use ark_std::UniformRand;
 pub use constant::ConstBasicBlock;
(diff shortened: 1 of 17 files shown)


ddkang:
Can we get some performance statistics on this?

src/basic_block.rs
```diff
@@ -50,6 +50,17 @@ pub struct SRS {
   pub Y2P: G2Projective,
 }
 
+// During proofs and verifications, a cache is used to prevent recomputation.
+// These are the types of the elements in the cache.
+pub enum CacheValues {
+  CQTableDict(HashMap<Fr, usize>),
+  CQ2TableDict(HashMap<(Fr, Fr), usize>),
+  Fr(Fr),
```
ddkang: Why do we need this?

arifiorino: There is a cache that the basic blocks use to prevent recomputation in proofs and verifications. This cache has strings as it's keys, and it is convenient if it can have multiple types as it's values. In order for the cache to have multiple types as it's values, and enum is used. Therefore the resulting cache has type `HashMap<String, CacheValues>`.

arifiorino:
Yeah, I'll work on getting numbers for Llama 2-7B

src/basic_block.rs
ddkang: Why is Fr needed?

arifiorino: It's needed so that alpha and beta can be shared across proofs. alpha and beta are random points used in the RLC proofs, and they are of type `Fr`.

ddkang: Why don't we call it AlphaBeta or like "RLCRandom" or something descriptive

arifiorino: Yeah, I'll change it to something more descriptive

liliatangxy (approved):

PR #40, opened 2024-05-24
Copy constraints for splatting ops

liliatangxy:
The existing protocol for `PermuteBasicBlock` cannot handle operations like window-based splatting needed for proving Conv and pooling ops where an element can appear in multiple positions and when an index for one dimension of the permutation depends on indices of more than one dimension.

The basic block itself requires that the input and output last dimensions are the same, meaning that in order to perform a splat, we have to perform a shape op to make the dimensions the same. We should be able to do this with any basic block that can prove padding. However, `self.permutation` is based on the indices and shapes of the original padded input/output.  Right now it expands the last index of `self.permutation` such that the elements of the original are evenly spaced, but I can clean this up to just pad trailing zeros.

This draft PR implements the copy constraint check based on the Plonk paper Section 8 protocol, but without quotient polynomial batching.


ddkang (commented):
The setup seems complicated enough it should be a struct

src/basic_block/copy_constraint.rs
```diff
@@ -0,0 +1,397 @@
...
+      product.push(product[d - 1] * shape[shape.ndim() - d]);
+    }
+  }
+  product = product.into_iter().rev().collect();
+  idx.as_array_view().iter().enumerate().fold(0, |sum, (i, x)| *x * product[i] + sum)
+}
+
+// If is_input, idxs is [(flat_idx of the input index, 0)]. Otherwise,
+// idxs is [(flat_idx of the output, flat_idx of the permuted input idx)]
+fn construct_ssig_coeffs(idxs: &[(usize, usize)], partitions: &HashMap<usize, Vec<usize>>, is_input: bool) -> Vec<Fr> {
```
ddkang: What is ssig?

src/basic_block/copy_constraint.rs
```diff
@@ -0,0 +1,397 @@
...
+    let input_len = input.first().unwrap().raw.len();
+    let output_len = output.first().unwrap().raw.len();
+
+    // Assumes that the input and output degrees are the same. To do this,
+    // we can permute the one with a smaller degree
+    assert!(input_len == output_len);
+    let N = input_len;
+    let domain = GeneralEvaluationDomain::<Fr>::new(N).unwrap();
+
+    let beta = Fr::rand(rng);
```
ddkang, line 183: Does this need to depend on anything? Can the prover cheat here?

src/basic_block/copy_constraint.rs
```diff
@@ -0,0 +1,415 @@
...
+  pub output_dim: IxDyn,
+}
+
+impl BasicBlock for CopyConstraintBasicBlock {
+  fn run(&self, _model: &ArrayD<Fr>, inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
+    assert!(inputs.len() == 1);
+    assert!(Dim(self.permutation.shape()) == self.output_dim);
+
+    // The basic block inputs and outputs will be scaled to the same degree,
+    // meaning that the last dimensions will be the same.
```
Vimalakirti: What's the reason to scale them to the same degree?

src/basic_block/copy_constraint.rs
```diff
@@ -0,0 +1,415 @@
...
+    let mut proof_0 = vec![L_i_x_1[0], sid_x];
+    proof_0.append(&mut ssig_polys.iter().map(|x| util::msm::<G1Projective>(&srs.X1A, &x.coeffs)).collect());
+
+    let mut proof_2 = vec![sid_poly];
+    proof_2.append(&mut ssig_polys);
+
+    return (proof_0, vec![L_i_x_2[0]], proof_2);
+  }
+
+  fn prove(
```
Vimalakirti, line 164: I feel this is exactly the same as the copy constraint described in Plonk. Is there anything I miss? (different from the paper?)

Vimalakirti (commented):
This PR implements the copy constraint (in Plonk).
Left some comments and also curious about Daniel's questions.
Please take a look at them, I'll review it carefully again later.

src/basic_block/copy_constraint.rs
```diff
@@ -0,0 +1,415 @@
...
+      vec![ssig]
+    })
+    .collect()
+}
+
+// Permutation has the same shape as the output, and each index stores the index of the input array it equals to
+#[derive(Debug)]
+pub struct CopyConstraintBasicBlock {
+  pub permutation: ArrayD<IxDyn>,
+  pub partitions: HashMap<usize, Vec<usize>>,
```
arifiorino: This variable can be moved to the `ProveVerifyCache`

src/tests.rs
```diff
@@ -2,35 +2,41 @@ use crate::basic_block::*;
...
 use ark_ec::pairing::{Pairing, PairingOutput};
+use ark_poly::univariate::DensePolynomial;
+use ark_std::One;
 use ark_std::UniformRand;
 use ark_std::Zero;
 use ndarray::{arr1, concatenate, s, ArrayD, Axis, IxDyn};
 use rand::{rngs::StdRng, SeedableRng};
 use std::collections::HashMap;
 
+use self::copy_constraint::CopyConstraintBasicBlock;
```
arifiorino: `CopyConstraintBasicBlock` should be added to the file basic_block.rs and then we can delete this line because all the basic blocks are imported on line 1 of this file.

src/basic_block/copy_constraint.rs
```diff
@@ -0,0 +1,415 @@
...
+    let N = max(self.input_dim[self.input_dim.ndim() - 1], self.output_dim[self.output_dim.ndim() - 1]);
+    let mut outp_shape = self.permutation.shape().to_vec().clone();
+    outp_shape[last_outp_dim - 1] = N;
+    vec![ArrayD::from_shape_fn(outp_shape, |i| {
+      let mut outp_idx = i.clone();
+      outp_idx[self.output_dim.ndim() - 1] = outp_idx[self.output_dim.ndim() - 1] * last_outp_dim / N;
+
+      if i[i.ndim() - 1] % (N / last_outp_dim) == 0 {
+        let mut inp_idx = self.permutation[&outp_idx].clone();
+        inp_idx[self.input_dim.ndim() - 1] = inp_idx[self.input_dim.ndim() - 1] * N / last_inp_dim;
```
arifiorino: These two lines are unnecessary. You define a new variable `inp_idx`, then modify it, and then never use it.

src/basic_block/copy_constraint.rs
```diff
@@ -0,0 +1,415 @@
...
+    assert!(inputs.len() == 1);
+    assert!(Dim(self.permutation.shape()) == self.output_dim);
+
+    // The basic block inputs and outputs will be scaled to the same degree,
+    // meaning that the last dimensions will be the same.
+    // permutation will correspond to the actual padded dimensions for the
+    // intput and output, so we need to adjust the input and output indices
+    // to correspond to permutation and fill in the output values that are
+    // not part of the actual permutation
+    let last_inp_dim = self.input_dim[self.input_dim.ndim() - 1];
```
arifiorino: This can be replaced with `self.input_dim.last().unwrap()`.

src/basic_block/copy_constraint.rs
```diff
@@ -0,0 +1,415 @@
...
+    let last_outp_dim = self.output_dim[self.output_dim.ndim() - 1];
+    let N = max(self.input_dim[self.input_dim.ndim() - 1], self.output_dim[self.output_dim.ndim() - 1]);
```
arifiorino: `tests.rs` doesn't have a test case for if these two values are different (it never tests for the padding). Can you add a test case for this?

arifiorino: Also, it seems that you pad the output if the input is smaller than the output. But what about the other case? What if the output is smaller than the input?

arifiorino (changes requested):
Can you write a Convolution `Layer`? Then it would be easy to plug it into the onnx framework.

src/basic_block/copy_constraint.rs
liliatangxy: Yeah it is the same protocol

src/basic_block/copy_constraint.rs
liliatangxy: It is unused so I removed it

liliatangxy:
Fixed some of the padding stuff and added tests for it.
It turns out that with the existing padding method we do not need to pad the inputs and outputs, we just need to pad the setup and permutation indices, so we no longer need PermuteBasicBlock before or after for dimension changes.

src/basic_block/copy_constraint.rs
liliatangxy: At this point no new proof elements have been made since the previous basic block, so I think we don't need to make Fiat-Shamir adjustments here.

src/basic_block/copy_constraint.rs
liliatangxy: I updated the comments for this function and `flat_index`. I can provide more detail if needed

src/basic_block/copy_constraint.rs
ddkang: can you add a comment?

src/basic_block/copy_constraint.rs
ddkang: I mean very literally, what is Sigma_j? Do you expect someone reading this file to know?

src/basic_block/copy_constraint.rs
liliatangxy: Doesn't seem to work for `IxDyn`

PR #41, opened 2024-05-24
Distilbert operations

arifiorino:
Adds the following `Layer`s that are in DistilBERT:
- Cast
- Div
- Equal
- Erf
- Expand
- Gather
- Mul
- Pow
- Reducemean
- Reshape
- Shape
- Softmax
- Sqrt
- Sub
- Transpose
- Where

A `Layer` now has access to the shape of it's inputs, as well as it's inputs that are constant.
If all the inputs of a `Layer` are constant, the `Layer` is run and it's output becomes constant.


src/basic_block/constant.rs
```diff
@@ -1,11 +1,29 @@
...
 use ndarray::ArrayD;
+use rand::rngs::StdRng;
 
 #[derive(Debug)]
 pub struct ConstBasicBlock;
 impl BasicBlock for ConstBasicBlock {
   fn run(&self, model: &ArrayD<Fr>, _inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
     vec![model.clone()]
   }
+  fn encodeOutputs(&self, _srs: &SRS, model: &ArrayD<Data>, _inputs: &Vec<&ArrayD<Data>>, _outputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Data>> {
```
ddkang, line 13: space

ddkang: Also what is this new function?

src/basic_block/div.rs
```diff
@@ -30,3 +30,18 @@ impl BasicBlock for DivScalarBasicBlock {
     vec![arr1(&div).into_dyn(), arr1(&rem).into_dyn()]
   }
 }
+
+#[derive(Debug)]
+pub struct DivConstBasicBlock {
+  pub c: f32,
+}
```
ddkang, line 37: space

src/basic_block/transpose.rs
```diff
@@ -0,0 +1,19 @@
+use super::{BasicBlock, Data, SRS};
+use ark_bn254::Fr;
+use ndarray::ArrayD;
+
+#[derive(Debug)]
+pub struct TransposeBasicBlock {
+  pub perm: Vec<usize>,
+}
+impl BasicBlock for TransposeBasicBlock {
```
ddkang, line 10: Spaces. Everywhere. I've also mentioned this SEVERAL times

ddkang (commented):
Please split out the reshape so we can merge this PR and Lilia's PR in some sequencing

src/basic_block/constant.rs
arifiorino: `encodeOutputs` was added to the `BasicBlock` trait in an earlier PR. I decided to implement it for `ConstBasicBlock` to speed things up. Since the model has already been encoded at this state, and the output equals the model, we can just clone the encoded model and return that as the encoded output. This saves us time because we don't encode things twice (which involves a lot of MSMs).

src/util.rs
```diff
@@ -253,3 +262,30 @@ pub fn combine_pairing_checks(checks: &Vec<&PairingCheck>) -> (Vec<G1Affine>, Ve
   println!("{:?}", res.0.len());
   res
 }
+
+pub fn broadcastDims(dims: &Vec<&Vec<usize>>, N: usize) -> Vec<usize> {
```
Vimalakirti, line 266: I think it's not obvious to see what N is for and how this function serves for. Maybe adding  a comment helps.

src/onnx.rs
```diff
@@ -1,17 +1,43 @@
...
 use crate::graph::*;
-use crate::layer;
-use crate::layer::Layer;
+use crate::layer::*;
+use crate::util;
 use ark_bn254::Fr;
 use ndarray::ArrayD;
 use std::collections::HashMap;
-use tract_onnx::prelude::Framework;
+use tract_onnx::prelude::{DatumType, Framework};
```
Vimalakirti, line 8: I'm curious could we use `tract_onnx` to load QOperator ONNX later?

src/layer/reshape.rs
```diff
@@ -0,0 +1,47 @@
...
+      let reshape = graph.addBB(Box::new(ReshapeBasicBlock { shape: intermediateShape }));
+      let permutation = ((0..a).map(|x| x * b).collect(), (0..b).collect());
+      let permute = graph.addBB(Box::new(RepeaterBasicBlock {
+        basic_block: Box::new(PermuteBasicBlock { permutation: permutation }),
+        N: 2,
+      }));
+      let intermediate = graph.addNode(reshape, vec![(-1, 0)]);
+      let output = graph.addNode(permute, vec![(intermediate, 0)]);
+      graph.outputs.push((output, 0));
+    } else {
```
Vimalakirti, line 45: what happened here (the else)?

src/layer/sqrt.rs
```diff
@@ -0,0 +1,23 @@
...
+use crate::graph::*;
+use crate::layer::Layer;
+use ark_bn254::Fr;
+use ndarray::ArrayD;
+
+pub struct SqrtLayer;
+impl Layer for SqrtLayer {
+  fn graph(input_shapes: &Vec<&Vec<usize>>, _constants: &Vec<Option<&ArrayD<Fr>>>) -> (Graph, Vec<Vec<usize>>) {
+    let mut graph = Graph::new();
+    let sqrt = graph.addBB(Box::new(SqrtBasicBlock { input_SF: 3, output_SF: 3 }));
```
Vimalakirti: SqrtLayer and ReLULayer are really similar. Why don't we implement a NonlinearLayer for them?

src/util.rs
Vimalakirti: And maybe manually choosing N is not the best way to implement broadcasting for ONNX (https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md)

Vimalakirti (approved):
This PR adds many new BBs and layers for Distilbert.
I'm confused about some parts such as the broadcasting mechanism.
But except for them, I found no big issue in this PR. Feel free to merge it.

src/util.rs
arifiorino: `N` depends on the properties of the underlying `basic_block`. If the underlying `basic_block` runs on 1d-arrays, `N=1` (for example `AdditionBasicBlock`). But if the underlying `basic_block` runs on 2d-arrays such as `MatMulBasicBlock`, then `N=2`.

src/onnx.rs
arifiorino: No, `tract_onnx` doesn't do quantization.

src/layer/reshape.rs
arifiorino: There are three types of reshaping that we support. The first case is when we aren't reshaping the last axis, in which case we can just run `ReshapeBasicBlock` (for example `8x2x4 -> 4x4x4`). In the second case, we split up the last axis into two axes (for example `8x4x4 -> 4x8x2x2`) And in the third case, we combine the last two axes into one (for example `4x8x2x2 -> 8x4x4`).

src/layer/sqrt.rs
arifiorino: Yeah I'm eventually gonna make a macro for this.

src/onnx.rs
Vimalakirti: Then how can we prove those ONNX in QOp?

src/util.rs
Vimalakirti: Could `N` be an attribute in `basic_block` so that we don't need to manually set `N`?

src/main.rs
```diff
@@ -98,11 +106,11 @@ fn verify(srs: &SRS, graph: &Graph) {
 }
 
 fn main() {
-  let srs = &ptau::load_file("challenge", 7);
-  let (mut graph, models) = onnx::load_file("sample.onnx");
+  let srs = &ptau::load_file("challenge14", 14, 14);
```
liliatangxy: This file needs to be added to the repo

src/basic_block/constant.rs
arifiorino: I added the space

src/basic_block/div.rs
arifiorino: I added the space

src/basic_block/transpose.rs
arifiorino: I added the space

src/onnx.rs
arifiorino: We would add layers to support that in the future.

src/util.rs
arifiorino: Yes, in the future we could add this. There might be a certain case where the same basicblock would require a different `N` though.

src/main.rs
arifiorino: I changed it back to use `challenge`.

src/layer/sqrt.rs
ddkang: Raise an issue

src/util.rs
ddkang: Raise an issue

src/graph.rs
```diff
@@ -25,8 +25,18 @@ impl Graph {
...
     self.nodes.iter().enumerate().for_each(|(i, n)| {
-      println!("running {i} {:?}", n.basic_block);
-      let myInputs = n.inputs.iter().map(|(j, k)| if *j < 0 { inputs[*k] } else { &(outputs[*j as usize][*k]) }).collect();
+      println!("running {i} {:?}", self.basic_blocks[n.basic_block]);
+      let myInputs = n
+        .inputs
+        .iter()
+        .map(|&(j, k)| {
+          if j < 0 {
+            inputs[k + (-j - 1) as usize]
```
liliatangxy, line 36: Is this making assumptions on the number of outputs of a layer or basic block?

src/layer/cast.rs
```diff
@@ -0,0 +1,17 @@
...
+use crate::layer::Layer;
+use ark_bn254::Fr;
+use ndarray::ArrayD;
+use tract_onnx::pb::AttributeProto;
+
+pub struct CastLayer;
+impl Layer for CastLayer {
+  fn graph(input_shapes: &Vec<&Vec<usize>>, _constants: &Vec<Option<&ArrayD<Fr>>>, _attributes: &Vec<&AttributeProto>) -> (Graph, Vec<Vec<usize>>) {
+    let mut graph = Graph::new();
+    let id = graph.addBB(Box::new(IdBasicBlock {}));
```
liliatangxy, line 12: Do we need to have a basic block here at all?

liliatangxy (approved):

src/layer/cast.rs
arifiorino: Yeah in order for the larger graph to run, we need some basic block here. But the identity basic block is extremely cheap.

src/graph.rs
arifiorino: I added a comment for this.

PR #42, opened 2024-05-28
Add concat, squeeze, and unsqueeze layers

Vimalakirti:
(as title states)
The Unsqueeze can only support input data from other nodes instead of itself. This will be fixed later.


src/basic_block/concat.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+    assert!(outputs[0].shape()[0] == inputs.len());
+    for i in 0..inputs.len() {
+      inputs[i].iter().zip(outputs[0].index_axis(Axis(self.axis), i).iter()).for_each(|(input, output)| {
+        assert!(input.g1 == output.g1);
+      });      
+    }
+    vec![]
+  }
+}
```
liliatangxy: please avoid removing newlines at the end of files

liliatangxy (approved):

src/basic_block/concat.rs
```diff
@@ -0,0 +1,58 @@
...
+use ndarray::{ArrayD, Axis};
+use rand::rngs::StdRng;
+
+// only support concat over dim 0 for now
+#[derive(Debug)]
+pub struct ConcatBasicBlock {
+  pub axis: usize,
+}
+
+impl BasicBlock for ConcatBasicBlock {
```
arifiorino, line 13: The logic here is wrong. This should only be defined for 1d vectors because otherwise this can be done with `ReshapeBasicBlock`. And for 1d vectors, this can be done in a manner similar to `PermuteBasicBlock`.

src/layer/squeeze.rs
```diff
@@ -0,0 +1,85 @@
...
+    let startShape = input_shapes[0];
+    let endShape: Vec<_> = (0..startShape.len() + 1).map(|x| if x == axis as usize { 1 } else { if x > axis as usize { startShape[x-1] } else { startShape[x] } }).collect();
+
+    if startShape.last() == endShape.last() {
+      let reshape = graph.addBB(Box::new(ReshapeBasicBlock { shape: endShape.clone() }));
+      let output = graph.addNode(reshape, vec![(-1, 0)]);
+      graph.outputs.push((output, 0));
+    } else if startShape.last() > endShape.last() {
+      let n = endShape.len();
+      let (mut a, mut b) = (endShape[n - 2], endShape[n - 1]);
```
arifiorino: `b` always equals 1

src/layer/squeeze.rs
```diff
@@ -0,0 +1,85 @@
...
+      assert!(*startShape.last().unwrap() == a * b);
+      let mut intermediateShape = endShape[..n - 2].to_vec();
+      intermediateShape.push(1);
+      intermediateShape.push(*startShape.last().unwrap());
+      intermediateShape.iter_mut().for_each(|x| *x = util::next_pow(*x as u32) as usize);
+      let reshape = graph.addBB(Box::new(ReshapeBasicBlock { shape: intermediateShape }));
+      (a, b) = (util::next_pow(a as u32) as usize, util::next_pow(b as u32) as usize);
+      let permutation = ((0..a).map(|x| x * b).collect(), (0..b).collect());
```
arifiorino: This can be rewritten given that `b` always equals 1

src/layer/squeeze.rs
```diff
@@ -0,0 +1,85 @@
...
+      let output = graph.addNode(reshape, vec![(intermediate, 0)]);
+      graph.outputs.push((output, 0));
+    }
+
+    (graph, vec![endShape])
+  }
+}
+
+pub struct UnsqueezeLayer;
+impl Layer for UnsqueezeLayer {
```
arifiorino, line 77: I just wanted to make clear that we're on the same page about the logic here, and maybe a comment would be useful. This layer adds an axis of length 1. If the axis is anything but the last axis, this can be done trivially using reshape. Otherwise, say we have a shape of `(a,b,c,d)`. Then we first want to insert an axis of length 1 here: `(a,b,c,1,d)`. Then we do a `PermuteBasicBlock` to get `(a,b,c,d,1)`.

src/layer/squeeze.rs
```diff
@@ -0,0 +1,85 @@
...
+    assert!(startShape[axis as usize] == 1);
+    let endShape: Vec<_> = startShape.iter().enumerate().filter(|(i, _)| *i != axis as usize).map(|(_, x)| *x).collect();
+    
+    if startShape.last() == endShape.last() {
+      let reshape = graph.addBB(Box::new(ReshapeBasicBlock { shape: endShape.clone() }));
+      let output = graph.addNode(reshape, vec![(-1, 0)]);
+      graph.outputs.push((output, 0));
+    } else { // startShape.last() < endShape.last()
+      let n = startShape.len();
+      let (mut a, mut b) = (startShape[n - 2], startShape[n - 1]);
```
arifiorino: `b` always equals 1

src/layer/squeeze.rs
```diff
@@ -0,0 +1,85 @@
...
+use ark_bn254::Fr;
+use ndarray::ArrayD;
+use tract_onnx::pb::AttributeProto;
+
+pub struct SqueezeLayer;
+impl Layer for SqueezeLayer {
+  fn graph(input_shapes: &Vec<&Vec<usize>>, _constants: &Vec<Option<&ArrayD<Fr>>>, attributes: &Vec<&AttributeProto>) -> (Graph, Vec<Vec<usize>>) {
+    let mut graph = Graph::new();
+
+    let axis: isize = attributes.iter().filter(|x| x.name == "axes").next().unwrap().ints[0] as isize;
```
arifiorino: Sometimes there are multiple axes which are supposed to be squeezed. Or sometimes there are no axes, and it is assumed that all axes of length 1 are to be removed.

src/layer/squeeze.rs
```diff
@@ -0,0 +1,85 @@
...
+    if startShape.last() == endShape.last() {
+      let reshape = graph.addBB(Box::new(ReshapeBasicBlock { shape: endShape.clone() }));
+      let output = graph.addNode(reshape, vec![(-1, 0)]);
+      graph.outputs.push((output, 0));
+    } else { // startShape.last() < endShape.last()
+      let n = startShape.len();
+      let (mut a, mut b) = (startShape[n - 2], startShape[n - 1]);
+      assert!(*endShape.last().unwrap() == a * b);
+      (a, b) = (util::next_pow(a as u32) as usize, util::next_pow(b as u32) as usize);
+      let permutation = (vec![0], (0..a * b).collect());
```
arifiorino: This can be rewritten given that `b` always equals 1

src/layer/concat.rs
```diff
@@ -0,0 +1,26 @@
+use crate::basic_block::*;
+use crate::graph::*;
+use crate::layer::Layer;
+use ark_bn254::Fr;
+use ndarray::ArrayD;
+use tract_onnx::pb::AttributeProto;
+
+pub struct ConcatLayer;
+impl Layer for ConcatLayer {
```
arifiorino, line 37: This should run `ReshapeBasicBlock` if the axis is anything except the last axis. Otherwise it should run `ConcatBasicBlock` (wrapped by a `RepeaterBasicBlock`) which is assumed to only run on 1d vectors.

arifiorino (changes requested):
I reviewed the files manually and found some mistakes. But I haven't run these layers myself to test, have you run the layers yourself?

Vimalakirti:
> I reviewed the files manually and found some mistakes. But I haven't run these layers myself to test, have you run the layers yourself?

yeah, I've manually created one onnx file with Squeeze, Unsqueeze, and Concat. This PR successfully proves the file.

src/basic_block/concat.rs
Vimalakirti: Why should it be defined for 1d vectors? I think the logic here works well for any concatenation except for the last dimension.

src/layer/squeeze.rs
Vimalakirti: thanks, I will simplify this part

src/layer/squeeze.rs
Vimalakirti: Thanks, let me fix it.

src/layer/concat.rs
Vimalakirti: I think the problem here is that `ReshapeBasicBlock` can only accept one input instead of `N` inputs for `Concat`. Thus, `ConcatBasicBlock` seems to be fine for me. Maybe creating one `ConcatLastDimBasicBlock` similar to `PermuteBasicBlock` can address your concern (?).

src/basic_block/concat.rs
Vimalakirti: Thanks, let me fix it.

src/basic_block/concat.rs
Vimalakirti: (There's a similar question below. I put a possible solution there.)

src/layer/concat.rs
arifiorino: Yes, this is true. Yes, that could work

src/basic_block/concat.rs
arifiorino: Yes, you are right. But also this file does a bunch of stuff with `Axis(0)` which is irrelevant. The only axis that is significant is the last axis, not the first one.

.gitignore
```diff
@@ -2,3 +2,8 @@ target
 Cargo.lock
 .DS_Store
 *.swp
+inputsEnc
+modelsEnc
+outputsEnc
+proofs
+*.onnx
```
arifiorino: I think we still want `sample.onnx` for testing purposes to still be in the repo.

src/basic_block/concat.rs
```diff
@@ -0,0 +1,56 @@
+use super::{BasicBlock, Data, DataEnc, PairingCheck, ProveVerifyCache, SRS};
+use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
+use ark_std::Zero;
+use ndarray::{Array1, ArrayD, Axis};
+use rand::rngs::StdRng;
+
+// only support concat over dim 0 for now
```
arifiorino: We could maybe support all axes except for the last axis.

src/basic_block/concat.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+impl BasicBlock for ConcatBasicBlock {
+  fn run(&self, _model: &ArrayD<Fr>, inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
+    for input in inputs.iter() {
+      assert!(inputs[0].shape()[1..] == input.shape()[1..]);
+    }
+    let mut r = inputs[0].clone().to_owned();
+    for input in inputs.iter().skip(1) {
+      r = ndarray::concatenate(Axis(self.axis), &[r.view(), input.view()]).unwrap();
```
arifiorino: We can combine all these calls into one call to `ndarray::concatenate`

src/basic_block/concat.rs
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  fn encodeOutputs(&self, _srs: &SRS, _model: &ArrayD<Data>, inputs: &Vec<&ArrayD<Data>>, _outputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Data>> {
+    let mut r = inputs[0].clone().to_owned();
+    if inputs[0].ndim() == 0 {
+      let r_vec = inputs.iter().map(|input| input.first().unwrap().clone()).collect::<Vec<Data>>();
+      r = Array1::from_vec(r_vec).into_dyn();
+    } else {
+      for input in inputs.iter().skip(1) {
+        r = ndarray::concatenate(Axis(self.axis), &[r.view(), input.view()]).unwrap();
```
arifiorino: We can combine all these calls into one call to `ndarray::concatenate`

src/basic_block/concat.rs
```diff
@@ -0,0 +1,56 @@
...
+    _srs: &SRS,
+    _model: &ArrayD<DataEnc>,
+    inputs: &Vec<&ArrayD<DataEnc>>,
+    outputs: &Vec<&ArrayD<DataEnc>>,
+    _proof: (&Vec<G1Affine>, &Vec<G2Affine>, &Vec<Fr>),
+    _rng: &mut StdRng,
+    _cache: &mut ProveVerifyCache,
+  ) -> Vec<PairingCheck> {
+    for i in 0..inputs.len() {
+      inputs[i].iter().zip(outputs[0].index_axis(Axis(self.axis), i).iter()).for_each(|(input, output)| {
```
arifiorino: The logic here is wrong, I don't think we should use `index_axis`. Here is a photo of how concatenate works which doesn't match this logic:
Screenshot 2024-06-09 at 7 06 31 PM

src/basic_block/concat.rs
```diff
@@ -0,0 +1,50 @@
...
+// support concat over any dim except for the last
+#[derive(Debug)]
+pub struct ConcatBasicBlock {
+  pub axis: usize,
+}
+
+impl BasicBlock for ConcatBasicBlock {
+  fn run(&self, _model: &ArrayD<Fr>, inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
+    for input in inputs.iter() {
+      assert!(inputs[0].shape()[1..] == input.shape()[1..]);
```
arifiorino: This assert is not right. Also we should assert that the axis is not the last axis.

src/basic_block/copy_constraint.rs
```diff
@@ -50,11 +54,17 @@ fn flat_index(shape: &IxDyn, idx: &IxDyn, N: usize) -> usize {
 // that should be the same over inputs and outputs.
 // If is_input, idxs is [(flat_idx of the input index, 0)]. Otherwise,
 // idxs is [(flat_idx of the output, flat_idx of the permuted input idx)]
-fn construct_ssig(idxs: &[(usize, usize)], N: usize, last_dim: usize, partitions: &HashMap<usize, Vec<usize>>, is_input: bool) -> Vec<Fr> {
+fn construct_ssig(
```
arifiorino: It seems that this PR is building off of #43. Therefore we should wait to push that PR before we push this one.

src/layer/concat.rs
```diff
@@ -0,0 +1,75 @@
+use crate::basic_block::*;
+use crate::graph::*;
+use crate::layer::Layer;
+use crate::util;
+use ark_bn254::Fr;
+use ndarray::{ArrayD, IxDyn};
+use tract_onnx::pb::AttributeProto;
+
+fn get_concat_indices(input_shapes: &Vec<&Vec<usize>>, output_shape: &Vec<usize>, axis: usize) -> Vec<ArrayD<Option<IxDyn>>> {
```
arifiorino, line 14: Can you add a comment explaining what this function does?

src/layer/concat.rs
```diff
@@ -0,0 +1,75 @@
...
+      }));
+
+      let mut cc_outputs = vec![];
+      for i in 0..input_shapes.len() {
+        let cc_output = graph.addNode(cc_basicblocks[i], vec![(-(i as i32 + 1), 0)]);
+        cc_outputs.push((cc_output, 0));
+      }
+      // add 2 cc_outputs and reduce to 1 output until only 1 output left
+      while cc_outputs.len() > 1 {
+        let add_output = graph.addNode(add, vec![cc_outputs.pop().unwrap(), cc_outputs.pop().unwrap()]);
```
arifiorino, line 73: This seems inefficient and it is also unsafe since `copy_constraint` is unsafe.

src/basic_block/copy_constraint.rs
Vimalakirti: yeah, let's wait for it.

src/layer/concat.rs
Vimalakirti: I think this cannot be more efficient as AddBasicBlock only accepts two inputs currently.
And `copy_constraint` will be safe soon after proper blinding.

arifiorino: No, this is not how this works. Don't resolve a conversation without actually resolving it.

Vimalakirti: Sorry, I thought my comment had answered your concerns. What's your suggestion then?

liliatangxy: Is it possible to combine each input to concat into a single `ArrayD` input? Then we would only need to use one `CopyConstraintBasicBlock` and no additions

Vimalakirti: I think it's not possible because the above case is for concatenation through the last dimension. We cannot generate a corresponding single ArrayD<Data> for proving in the case.
For instance, imaging 4 input arrays are in shape [16,2], then their commitment will all be in shape [16,1]. The output array would be in shape [16,8], of which commitment shape would be [16,1]. If we combine each input into single input of size [16,8] first for copy constraint, we do not know how to combine their commitments (4 [16,1] arrays) into 1 [16,1] Data array for copy constraint proving and verification.

src/layer/concat.rs
```diff
@@ -0,0 +1,81 @@
...
+    });
+    axis_offset += input_shapes[i][axis];
+    indices.push(output);
+  }
+  indices
+}
+
+pub struct ConcatLayer;
+impl Layer for ConcatLayer {
+  fn graph(input_shapes: &Vec<&Vec<usize>>, _constants: &Vec<Option<&ArrayD<Fr>>>, attributes: &Vec<&AttributeProto>) -> (Graph, Vec<Vec<usize>>) {
```
ddkang, line 38: Please leave comments describing the high level algorithm

src/layer/squeeze.rs
```diff
@@ -0,0 +1,122 @@
...
+use crate::graph::*;
+use crate::layer::Layer;
+use crate::util;
+use ark_bn254::Fr;
+use ndarray::{ArrayD, Axis};
+use tract_onnx::pb::AttributeProto;
+
+pub struct SqueezeLayer;
+impl Layer for SqueezeLayer {
+  fn graph(input_shapes: &Vec<&Vec<usize>>, _constants: &Vec<Option<&ArrayD<Fr>>>, attributes: &Vec<&AttributeProto>) -> (Graph, Vec<Vec<usize>>) {
```
ddkang, line 15: Ditto on comments

src/layer/concat.rs
ddkang: Is this comment resolved?

Vimalakirti: @liliatangxy @arifiorino

liliatangxy: lgtm

ddkang (approved):

PR #43, opened 2024-05-29
Conv layer

liliatangxy:
Added Conv layer. For now the layer proves weights splatting and uses MatMul instead of CQLin since fixing this would require changes to onnx processing.


ddkang:
Isn't the verifier check required for soundness?

src/layer/conv.rs
```diff
@@ -0,0 +1,191 @@
+use crate::basic_block::*;
```
ddkang, line 1: Unless this is a fully general ND conv, call it conv_<whatever>d

src/layer/conv.rs
```diff
@@ -0,0 +1,191 @@
...
+use crate::graph::*;
```
ddkang, line 2: This file is very complicated. Are there ways to simplify?

src/layer/conv.rs
liliatangxy: It is general

src/layer/conv.rs
```diff
@@ -0,0 +1,191 @@
...
+    let weight_shape = input_shapes[1];
+    let dims = input_shapes[0][2..].to_vec();
+    let ch_dims = weight_shape[2..].to_vec();
+
+    let strides: Vec<_> = attributes.iter().filter(|x| x.name == "strides").next().unwrap().ints.iter().map(|x| *x as usize).collect();
+    let pads: Vec<_> = attributes.iter().filter(|x| x.name == "pads").next().unwrap().ints.iter().map(|x| *x as usize).collect();
+
+    let permutation = splat_input(&input_shapes[0], &strides, &pads, weight_shape[1], &ch_dims);
+    let permutation_padded = splat_pad(&permutation);
+    let output_dim = permutation_padded.dim();
```
Vimalakirti: To prevent possible bugs, maybe we can remove `output_dim` from `CopyConstraintBasicBlock` as it is exactly the same as `permutation.dim()`

src/main.rs
```diff
@@ -108,11 +108,11 @@ fn verify(srs: &SRS, graph: &Graph) {
 }
 
 fn main() {
-  let srs = &ptau::load_file("challenge", 7, 7);
-  let (mut graph, models) = onnx::load_file("sample.onnx");
+  let srs = &ptau::load_file("challenge14", 14, 14);
```
Vimalakirti: This file is missing.

src/layer/conv.rs
```diff
@@ -0,0 +1,191 @@
...
+
+pub struct ConvLayer;
+impl Layer for ConvLayer {
+  fn graph(input_shapes: &Vec<&Vec<usize>>, _constants: &Vec<Option<&ArrayD<Fr>>>, attributes: &Vec<&AttributeProto>) -> (Graph, Vec<Vec<usize>>) {
+    let mut graph = Graph::new();
+    let weight_shape = input_shapes[1];
+    let dims = input_shapes[0][2..].to_vec();
+    let ch_dims = weight_shape[2..].to_vec();
+
+    let strides: Vec<_> = attributes.iter().filter(|x| x.name == "strides").next().unwrap().ints.iter().map(|x| *x as usize).collect();
```
Vimalakirti: "strides" may not exist sometimes (the default value is 1)

Vimalakirti (approved):

src/basic_block/copy_constraint.rs
```diff
@@ -295,17 +356,18 @@ impl BasicBlock for CopyConstraintBasicBlock {
...
     ];
     proof.append(&mut C);
-    let mut s_1s = setup.0[1..].iter().map(|x| Into::<G1Projective>::into(*x)).collect();
+    let mut s_1s = setup.0[2..].iter().map(|x| Into::<G1Projective>::into(*x)).collect();
     proof.append(&mut s_1s);
 
     let mut evals = vec![Z_ga];
     evals.append(&mut q1_evals);
-    return (proof, vec![setup.1[0].into()], evals);
+    return (proof, vec![setup.1[0].into(), setup.1[1].into()], evals);
```
arifiorino: This reveals the inputs to the verifier. Specifically, `fj_poly` are the polynomials of the inputs. And `q1_evals` includes those polynomials evaluated at a random challenge `a`. And then `q1_evals` is sent to the verifier since it is added to `evals`. Through a polynomial interpolation attack, the inputs can be recovered by the verifier.

arifiorino (changes requested):

src/layer/conv.rs
```diff
@@ -0,0 +1,212 @@
...
+use ndarray::indices;
+use ndarray::ArrayD;
+use ndarray::Axis;
+use ndarray::Dim;
+use ndarray::Dimension;
+use ndarray::IxDyn;
+use ndarray::Slice;
+use tract_onnx::pb::AttributeProto;
+
+pub fn pad<G: Clone>(input: &ArrayD<G>, padding: &Vec<[usize; 2]>, pad_val: &G) -> ArrayD<G> {
```
Vimalakirti: we have another `pub fn pad` in `util`. Maybe it would be clearer to add comments or change its name.

src/layer/conv.rs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+pub struct ConvLayer;
+impl Layer for ConvLayer {
+  fn graph(input_shapes: &Vec<&Vec<usize>>, _constants: &Vec<Option<&ArrayD<Fr>>>, attributes: &Vec<&AttributeProto>) -> (Graph, Vec<Vec<usize>>) {
+    let mut graph = Graph::new();
+    let weight_shape = input_shapes[1];
+    let dims = input_shapes[0][2..].to_vec();
+    let ch_dims = weight_shape[2..].to_vec();
+
```
Vimalakirti, line 102: Do we need to consider dilations?
(reference: https://onnx.ai/onnx/operators/onnx__Conv.html)

liliatangxy: I don't think any of the CNNs we are proving have non-default values for dilations. I can make a not implemented error for it

src/layer/conv.rs
liliatangxy: renamed the one in util since it is specific to padding to the next pow of 2. we could also move this to util

ddkang: Can these pad functions be merged?

src/layer/conv.rs
ddkang: Checking in on this.

liliatangxy: I think we can move `pad` to `utils`, merge it with the other pad, and get `splat_pad` to use the other pad.

ddkang: Sounds good

PR #46, opened 2024-06-14
Debug tool for verification process

Vimalakirti:
In our current verification process, if an ONNX model passes all `running`, `setting`, `encoding`, and `proving` stages but fails during `verifying`, we have no way of knowing which BasicBlock went wrong. This small PR adds a tool, `Graph.verify_for_each_pairing`, as an alternative to `Graph.verify` to aid in debugging.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -135,7 +135,9 @@ impl Graph {
     assert_eq!(Bn254::multi_pairing(pairings.0.iter(), pairings.1.iter()), PairingOutput::zero());
   }
 
-  pub fn verify_for_debug(
+  // This function should be only used for debugging purposes (it is very slow). 
+  // It verifies the proofs without combining pairing checks so that we can see which BasicBlock is failing.
+  pub fn verify_for_each_pairing(
     &self,
     srs: &SRS,
     models: &Vec<&ArrayD<DataEnc>>,
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -1,4 +1,3 @@
-use crate::basic_block;
 use crate::basic_block::*;
 use crate::util;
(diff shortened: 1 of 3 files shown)


src/graph.rs
```diff
@@ -134,6 +135,35 @@ impl Graph {
     assert_eq!(Bn254::multi_pairing(pairings.0.iter(), pairings.1.iter()), PairingOutput::zero());
   }
 
+  pub fn verify_for_debug(
```
ddkang: I don't know what this does from the name

src/graph.rs
```diff
@@ -134,6 +135,35 @@ impl Graph {
...
+    models: &Vec<&ArrayD<DataEnc>>,
+    inputs: &Vec<&ArrayD<DataEnc>>,
+    outputs: &Vec<&Vec<&ArrayD<DataEnc>>>,
+    proofs: &Vec<(&Vec<G1Affine>, &Vec<G2Affine>, &Vec<Fr>)>,
+    rng: &mut StdRng,
+  ) {
+    let mut cache = HashMap::new();
+    self.nodes.iter().enumerate().for_each(|(i, n)| {
+      println!("verifying (debug mode) {i} {:?}", self.basic_blocks[n.basic_block]);
+      let myInputs = n.inputs.iter().map(|(j, k)| if *j < 0 { inputs[*k] } else { &(outputs[*j as usize][*k]) }).collect();
```
ddkang: I think you can use a more descriptive variable here

src/graph.rs
Vimalakirti: Thanks, I've updated the fn name to `verify_for_each_pairing` and add comments to note it's only for debugging.

ddkang (approved):

liliatangxy (approved):

Issue #45, opened 2024-06-10
Missing blinding and optimizations in copy constraint


liliatangxy:
#40 introduced copy constraints that are currently insecure due to missing blinding factors. 
We can also batch the pairings with quotients at the end.

liliatangxy:
To completely fix the issue, it requires us to make changes to the Data and DataEnc constructors to address this case. As a workaround we can pass the correct blinding to the verifier so that we can support a validly blinded copy constraint, but to make it fully secure we need to support opening blinding in the `Data`/`DataEnc::new` paths.

ddkang:
> To completely fix the issue, it requires us to make changes to the Data and DataEnc constructors to address this case. As a workaround we can pass the correct blinding to the verifier so that we can support a validly blinded copy constraint, but to make it fully secure we need to support opening blinding in the `Data`/`DataEnc::new` paths.

Comment it in code

liliatangxy:
Closed with #47

PR #47, opened 2024-06-17
Copy constraint padding and blinding

liliatangxy:
Includes the padding support from #43 and implements the missing blinding for the input, output, and Z polynomials from #45. 

In order to do the latter, we change the terms in input, output, Z, t, r polynomials to match those found in Section 8 of the Plonk paper, but for now t and r only contain terms relevant to the f'(x)Z(x) = g'(x)(Zwx) check and will match the paper once we fully implement quotient batching. Batching will be completed in another commit or PR.

Would appreciate suggestions for tests based on other shape operations we need.

### DIFF
--- a/src/basic_block/copy_constraint.rs
+++ b/src/basic_block/copy_constraint.rs
@@ -25,23 +25,30 @@ use std::{
   iter::{once, repeat},
 };
 
-fn flat_index(shape: &IxDyn, idx: &IxDyn, N: usize) -> usize {
-  assert!(shape.ndim() == idx.ndim());
+fn flat_index(shape: &IxDyn, idx: &Option<IxDyn>, N: usize) -> Option<(usize, usize)> {
+  assert!(*idx == None || shape.ndim() == idx.as_ref().unwrap().ndim());
   let mut product = vec![];
   // If inputs and outputs do not have the same last dimension, then the one
   // with the smaller dimension will have had their polynomials constructed from
   // a smaller evaluation domain. This indexing enables the smaller dimension's
   // roots of unity evaluation values to line up to the larger one.
-  let spacing = N / shape[shape.ndim() - 1];
-  for d in 0..shape.ndim() {
-    if d == 0 {
-      product.push(spacing);
-    } else {
-      product.push(product[d - 1] * shape[shape.ndim() - d]);
+  if let Some(j) = idx {
+    let spacing = N / shape[shape.ndim() - 1];
+    product.push(1);
+    for d in 1..(shape.ndim() - 1) {
+      product.push(product[d - 1] * shape[shape.ndim() - 1 - d]);
     }
+    product = product.into_iter().rev().collect();
+    let left_idx = if shape.ndim() == 1 {
+      0
+    } else {
+      product.iter().enumerate().fold(0, |sum, (i, x)| *x * j[i] + sum)
+    };
+    let right_idx = j[shape.ndim() - 1] * spacing;
+    Some((left_idx, right_idx))
+  } else {
+    None
   }
-  product = product.into_iter().rev().collect();
-  idx.as_array_view().iter().enumerate().fold(0, |sum, (i, x)| *x * product[i] + sum)
 }
 
 // Construct S_sigma_j polynomial evals over N-th roots of unity
@@ -50,11 +57,18 @@ fn flat_index(shape: &IxDyn, idx: &IxDyn, N: usize) -> usize {
 // that should be the same over inputs and outputs.
 // If is_input, idxs is [(flat_idx of the input index, 0)]. Otherwise,
 // idxs is [(flat_idx of the output, flat_idx of the permuted input idx)]
-fn construct_ssig(idxs: &[(usize, usize)], N: usize, last_dim: usize, partitions: &HashMap<usize, Vec<usize>>, is_input: bool) -> Vec<Fr> {
+fn construct_ssig(
+  idxs: &[((usize, usize), Option<(usize, usize)>)],
+  N: usize,
+  last_dim: usize,
+  partitions: &HashMap<Option<(usize, usize)>, Vec<(usize, usize)>>,
+  is_input: bool,
+) -> Vec<Fr> {
+  let domain = GeneralEvaluationDomain::<Fr>::new(N).unwrap();
   idxs
     .iter()
     .flat_map(|(idx, perm_idx)| {
-      let inp_idx = if is_input { idx } else { perm_idx };
+      let inp_idx = if is_input { Some(*idx) } else { *perm_idx };
       let sigma = {
         let partition = partitions.get(&inp_idx).ok_or_else(|| format!("Key {:?} not found in the partition", inp_idx)).unwrap();
         if let Some(pos) = partition.iter().position(|x| *x == *idx) {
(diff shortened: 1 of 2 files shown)


src/basic_block/copy_constraint.rs
```diff
@@ -353,50 +453,56 @@ impl BasicBlock for CopyConstraintBasicBlock {
...
+        ]);
+        break;
+      }
+    }
 
-    let f1_a: Fr = fj1_as.iter().product();
-    let g1_a: Fr = gj1_as[..gj1_as.len() - 1].iter().product();
+    // Check Z(x)f'(x) = Z(gx)g'(x)
+    let ft_as: Vec<_> = fj_as.iter().enumerate().map(|(i, x)| beta * Fr::from((i + 1) as i32) * a + *x + gamma).collect();
+    let gt_as: Vec<_> = fj_as.iter().enumerate().map(|(i, x)| ssig_as[i] * beta + *x + gamma).collect();
```
Vimalakirti: Verification still fails: program crashes at this line when I try the Slice example I gave you. Error is "index out of bounds".

liliatangxy: Resolved, it was an issue with the graph indexing

Vimalakirti (approved):
I guess it will be really difficult for someone less familiar with Plonk permutation to understand what's going on here. While I know @liliatangxy has made a great effort to ensure every var name aligns with the ones in Plonk paper, maybe it helps to include the link of the paper somewhere and add page numbers after comments for others to refer to (e.g., section for T polynomial).

ddkang:
> I guess it will be really difficult for someone less familiar with Plonk permutation to understand what's going on here. While I know @liliatangxy has made a great effort to ensure every var name aligns with the ones in Plonk paper, maybe it helps to include the link of the paper somewhere and add page numbers after comments for others to refer to (e.g., section for T polynomial).

Good point. Please link to the relevant papers and sections @liliatangxy

ddkang:
Need to reduce pairings before being able to be merged

ddkang:
Does the padding need to be in copy or can they be made two separate blocks without performance decreases?

liliatangxy:
Reduced the verifier check pairings down to 3. The third one is due to us using the old blinding scheme for the t polynomial commitment (p. 29 in plonk paper). We use the old scheme because the blinding scheme for t in the paper relies on splitting the t commitment into multiple commitments over smaller deg polynomials. The reason to not split t up is that in the Plonk paper they only have to deal with 3 permutation polynomials, but in our case the number of polynomials we have is the size of the input and output data object, which makes it costly for the verifier (seen in the last line of D on p. 31).

Splitting out padding from this BB would have high verifier overheads. To prove padding, we have to prove that the padded parts are 0. This can either be done by 1) zero-checking each pad element, or by 2) zero-checking one of the elements and a permutation check with a cycle.

1 requires two additional msms the size of the degree of the Data poly in the prover. One for the batched quotients of the zero checks `L_mask (x) f(x) = 0` for each input/output poly containing padding and one commitment for openings. This is expensive for the verifier because we would have to do an msm to add up the products of the mask poly openings with the input poly commitments for each poly containing padding (RLC over `L_mask(a) [f]_1`s).
2 is basically redoing the permutation arg for just padding, so there's no reason to consider it.

src/basic_block/copy_constraint.rs
```diff
@@ -173,139 +213,194 @@ impl BasicBlock for CopyConstraintBasicBlock {
...
-    proof.serialize_uncompressed(&mut bytes).unwrap();
+    let r = Fr::rand(&mut rng2);
+    let mut proof_1 = vec![t_x + srs.Y1P * r];
+    proof_1.serialize_uncompressed(&mut bytes).unwrap();
     util::add_randomness(rng, bytes);
+    proof.append(&mut proof_1);
 
-    // Calculate opening argument for Z over omega * a
-    let omega = domain.group_gen();
     let a = Fr::rand(rng);
```
Vimalakirti: I think this should be `z`, described in p.29

src/basic_block/copy_constraint.rs
```diff
@@ -173,139 +213,194 @@ impl BasicBlock for CopyConstraintBasicBlock {
...
+    let mut fj_as: Vec<_> = fj_polys.iter().map(|p| p.evaluate(&a)).collect();
+
+    // Calculate opening argument for Z over omega * a (p. 30 of [1])
```
Vimalakirti: p.29?

src/basic_block/copy_constraint.rs
```diff
@@ -173,139 +213,194 @@ impl BasicBlock for CopyConstraintBasicBlock {
...
       coeffs: vec![-a * omega, Fr::one()],
     };
     let temp = Z_poly.sub(&Z_ga_poly);
     let Z_Q: DensePolynomial<_> = &temp / &Z_V;
-    let Z_Q_x = util::msm::<G1Projective>(&srs.X1A, &Z_Q.coeffs);
-
-    // Calculate opening quotient for Z(x)f'(x) = Z(gx)g'(x) check
-    let fj_poly_iter = fj_polys.iter().map(|x| *x);
-    let mut q1_evals: Vec<Fr> = once(sid_poly).chain(ssig_polys.iter()).chain(fj_poly_iter.clone()).map(|p| p.evaluate(&a)).collect();
+    let W_gx = util::msm::<G1Projective>(&srs.X1A, &Z_Q.coeffs);
```
Vimalakirti, line 373: I think it's `W_z`, and the comment for "Round 5" is missing

Vimalakirti (commented):
Thanks for the updates! There are few variable names/page numbers seem to be mismatched with the ones in Plonk. I didn't point out all of them, please check.

src/basic_block/copy_constraint.rs
liliatangxy: I thought z might be confusing with the Z polynomial, so I used a. Using the full name zeta might make other var names too long. I think it should be clear that a is for openings with the next comments

src/basic_block/copy_constraint.rs
liliatangxy: No, it refers to the W_zeta*omega commitment in round 5, updated the comment

src/basic_block/copy_constraint.rs
liliatangxy: Since we don't use zeta in the code, I thought using gx would be clearer since it corresponds with using g for omega/generator and x to indicate that it is after the msm.

Fixed fiat shamir round 5 openings to include the evaluations

liliatangxy:
The bug mentioned in the meeting with copy constraint didn't have anything to do with the input/output shapes. The issue was that the verifier needs to generate an additional random value that the prover doesn't need and I didn't have the prover generate it as well, so in subsequent basic blocks requiring randomness, they would generate different values. Fixed by having the prover generate that value as well.

Because the random values will only become offset in subsequent basic blocks, this issue isn't easily testable using the basic block unit tests. Not sure if we should design another test for it.

ddkang:
We should absolutely have tests for that scenario

liliatangxy:
Added a check for rng state at the end of `testBasicBlock` for this

src/basic_block/copy_constraint.rs
ddkang: can you actually just use the greek letters?

src/basic_block/copy_constraint.rs
ddkang: ditto with greek letters

ddkang (approved):

PR #49, opened 2024-06-17
Refactor onnx

Vimalakirti:
This PR refactors `onnx.rs`, adding the following features:
- Improved readability
- Ignored optional input fields of ONNX layers
- Computed precomputable constants without padding (which might cause problems for layers like `Concat`)
- Returned `fake_input` to facilitate testing the Graph (this tool spares us the effort of creating inputs for various ONNX files)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,7 +6,7 @@ use ark_poly::univariate::DensePolynomial;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use basic_block::*;
 use ndarray::ArrayD;
-use rand::{rngs::StdRng, Rng, SeedableRng};
+use rand::{rngs::StdRng, SeedableRng};
 use rayon::prelude::*;
 use sha3::{Digest, Keccak256};
 use std::fs::{self, File};
@@ -109,11 +109,8 @@ fn verify(srs: &SRS, graph: &Graph) {
 
 fn main() {
   let srs = &ptau::load_file("challenge", 7, 7);
-  let (mut graph, models) = onnx::load_file("sample.onnx");
-  let mut rng = StdRng::from_entropy();
-  let input: Vec<Fr> = (0..4).map(|_| Fr::from(rng.gen_range(-4..4))).collect();
-  let input = ArrayD::from_shape_vec(vec![1, 4], input).unwrap();
-  let inputs = vec![&input];
+  let (mut graph, models, fake_inputs) = onnx::load_file("sample.onnx");
+  let inputs = fake_inputs.iter().map(|x| x).collect();
   let models = models.iter().map(|x| x).collect();
   prove(&srs, &inputs, &mut graph, &models);
   verify(&srs, &graph);
--- a/src/main.rs
+++ b/src/main.rs
@@ -109,7 +109,9 @@ fn verify(srs: &SRS, graph: &Graph) {
 
 fn main() {
   let srs = &ptau::load_file("challenge", 7, 7);
-  let (mut graph, models, fake_inputs) = onnx::load_file("sample.onnx");
+  let onnx_file_name = "sample.onnx";
+  let (mut graph, models) = onnx::load_file(onnx_file_name);
+  let fake_inputs = util::generate_fake_inputs_for_onnx(onnx_file_name);
   let inputs = fake_inputs.iter().map(|x| x).collect();
   let models = models.iter().map(|x| x).collect();
   prove(&srs, &inputs, &mut graph, &models);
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -109,11 +109,8 @@ fn verify(srs: &SRS, graph: &Graph) {
 
 fn main() {
   let srs = &ptau::load_file("challenge", 7, 7);
-  let (mut graph, models) = onnx::load_file("sample.onnx");
-  let mut rng = StdRng::from_entropy();
-  let input: Vec<Fr> = (0..4).map(|_| Fr::from(rng.gen_range(-4..4))).collect();
-  let input = ArrayD::from_shape_vec(vec![1, 4], input).unwrap();
-  let inputs = vec![&input];
+  let (mut graph, models, fake_inputs) = onnx::load_file("sample.onnx");
```
ddkang: What exactly is fake_inputs and what is the contract?

src/onnx.rs
```diff
@@ -81,108 +86,240 @@ pub fn load_file(filename: &str) -> (Graph, Vec<ArrayD<Fr>>) {
...
-    if my_constants.iter().all(|&x| x.is_some()) {
-      let my_inputs: Vec<_> = my_constants.iter().map(|&x| x.unwrap().clone()).collect();
-      let my_inputs = my_inputs.iter().map(|x| x).collect();
-      let outputs = local_graph.run(&my_inputs, &vec![&arr1(&[]).into_dyn(); local_graph.basic_blocks.len()]);
-      node.output.iter().zip(local_graph.outputs.iter()).for_each(|(output_str, &(nodeX, nodeY))| {
-        passed_constants.insert(output_str, outputs[nodeX as usize][nodeY].clone());
-      });
-    }
+// This function is used for generating fake inputs for onnx models
+// It is only for testing purposes
```
ddkang: See above about the contracts

src/main.rs
Vimalakirti: `fake_inputs` are not real inputs read from a dataset. Instead, they are tensors filled with random elements, whose shapes match those described in the input tensors of an ONNX model. Generating these when loading an ONNX file saves us from creating different input tensors ourselves. About the contract, they should have the same shape and data type as the ONNX model's input tensors and should be consistently usable for testing and validation purposes within our system.
I'll update the script and add comments to make it clear.

Vimalakirti:
The new commit adds the following features:
- fixed indexing bugs in `graph.rs` when feeding `inputs` for `encodeOutputs`, `prove`, and `verify`
- added comments for `fake_inputs`
- renamed confusing variable names

ddkang:
@iantimmis we're beginning to refactor the ONNX parsing. It would be great if you could check to see if this improves readability!

src/main.rs
liliatangxy: I don't think we should include input generation as part of the core onnx file loading. Instead, we can separate this out into a utility function or function that is just used for testing purposes to do the same.

We might want to determine a way to make it easier for a developer to add their own inputs without needing to guess the input order.

src/onnx.rs
```diff
@@ -81,108 +87,252 @@ pub fn load_file(filename: &str) -> (Graph, Vec<ArrayD<Fr>>) {
...
+    "Where" => Ok(WhereLayer::graph(&input_shapes, &node_constants, &node_attributes)),
+    "Expand" => Ok(ExpandLayer::graph(&input_shapes, &node_constants, &node_attributes)),
+    "Softmax" => Ok(SoftmaxLayer::graph(&input_shapes, &node_constants, &node_attributes)),
+    "Erf" => Ok(ErfLayer::graph(&input_shapes, &node_constants, &node_attributes)),
+    _ => Err(format!("Unsupported onnx operation: {op}")),
+  }
+  .unwrap()
+}
+
+// This function is used for updating the graph of onnx after creating the local graph
```
liliatangxy: Can you add more comments so that we know what the intended logic of the operation as well as what the arguments are referring to?

src/onnx.rs
```diff
@@ -81,108 +87,252 @@ pub fn load_file(filename: &str) -> (Graph, Vec<ArrayD<Fr>>) {
...
+    });
+  }
+  for (i, output) in node.output.iter().enumerate() {
+    let local_output = local_graph.outputs[i];
+    outputs_idx.insert(output.clone(), vec![(start_idx + local_output.0, local_output.1)]);
+  }
+}
+
+// This function is used for processing a layer of onnx models
+fn process_node(
```
liliatangxy, line 207: More comments for this fn too would be nice

liliatangxy (approved):

PR #51, opened 2024-07-03
add gpu feature

Vimalakirti:
This PR introduces the gpu feature into zk-torch, which can allow us to zk-prove ML models using cuda gpus.
Currently, it's based on icicle and Rust conditional compilation. 
This way of implementation leaves us two issues:
- There are repeated codes due to the reason that we need two versions of `prove` (one for cpu and one for gpu) for each basicblock. I'll check if there exists a better way to address this issue.
- One needs to compile icicle and change the icicle path in `Cargo.toml` before using zk-torch.

(Note. a different platform may require a different way to compile icicle. Currently, I do have a ReadMe.md file to guide one to compile icicle on UIUC campus cluster. But we may also need to write guidelines for different platforms later.)


liliatangxy:
Is it possible to have wrapper functions for each of the calls that differ between cpu and gpu, and in the implementation of each wrapper we have the code depend on feature? 

Something like this:
```rust
#[cfg(feature = "gpu")]
pub fn wrapper_function(arg: SomeType) -> SomeReturnType {
    // ...
}

#[cfg(not(feature = "gpu"))]
pub fn wrapper_function(arg: SomeType) -> SomeReturnType {
    // ...
}
```

ddkang:
The automated checks are failing

src/basic_block.rs
```diff
@@ -12,6 +12,8 @@ pub use cq2::CQ2BasicBlock;
 pub use cqlin::CQLinBasicBlock;
 pub use div::{DivConstBasicBlock, DivScalarBasicBlock};
 pub use eq::EqBasicBlock;
+#[cfg(feature = "gpu")]
```
ddkang: Why can't these be collected in a single area?

src/basic_block.rs
```diff
@@ -69,7 +91,10 @@ pub enum CacheValues {
   Data(Data),
   G2(G2Affine),
 }
+#[cfg(not(feature = "gpu"))]
```
ddkang: why does this need to be different?

src/basic_block/add.rs
```diff
@@ -39,7 +39,8 @@ impl BasicBlock for AddBasicBlock {
     outputs: &Vec<&ArrayD<DataEnc>>,
     _proof: (&Vec<G1Affine>, &Vec<G2Affine>, &Vec<Fr>),
     _rng: &mut StdRng,
-    _cache: &mut ProveVerifyCache,
+    #[cfg(not(feature = "gpu"))] _cache: &mut ProveVerifyCache,
```
ddkang: Why do these need to be different?

ddkang (commented):
Having two implementations everywhere is unusably bad. Why can't we do what Lilia suggested?

Vimalakirti:
Thanks for all the valuable comments. The PR should be much clearer to read now. In the mean time, I will parallelize more things.

src/basic_block/cq2.rs
```diff
@@ -14,6 +14,75 @@ use rand::{rngs::StdRng, SeedableRng};
 use rayon::prelude::*;
 use std::collections::HashMap;
 
+#[cfg(not(feature = "gpu"))]
+fn cq2_setup_wrapper(srs: &SRS, model: &ArrayD<Data>) -> (Vec<G1Projective>, Vec<G2Projective>, Vec<DensePolynomial<Fr>>) {
```
ddkang: Why do we need two separate functions here?

ddkang: Why can't we simply call the helper functions in utils the same thing so that these functions have identical implementations?

src/basic_block/div.rs
```diff
@@ -39,10 +57,21 @@ pub struct DivConstBasicBlock {
...
     assert!(inputs.len() == 1);
-    vec![inputs[0].map(|x| {
+    #[cfg(not(feature = "gpu"))]
+    let out = inputs[0].map(|x| {
       let mut x = util::fr_to_int(*x) as f32;
       x /= self.c;
       Fr::from(x.round() as i32)
-    })]
+    });
+    #[cfg(feature = "gpu")]
```
ddkang: Can you make these feature flag blocks?

src/basic_block.rs
```diff
@@ -49,6 +54,21 @@ pub mod sub;
...
+  pub X2P: Vec<G2Projective>,
+  pub Y1A: G1Affine,
+  pub Y2A: G2Affine,
+  pub Y1P: G1Projective,
+  pub Y2P: G2Projective,
+  pub IX1A: Vec<IG1A>,
+  pub IX2A: Vec<IG2A>,
+}
+
+#[cfg(not(feature = "gpu"))]
```
ddkang: Can you do an if/else for these?

src/basic_block/cq2.rs
liliatangxy: Yeah I was thinking that we can have a `util::fft`, etc. which has a gpu and not gpu implementation. We can do the same for any operation that is gpu-optimized.

src/basic_block.rs
Vimalakirti: It no longer differs.

Vimalakirti:
Thanks again for the valuable comments. The PR is now ready for reviews.

src/util.rs
```diff
@@ -82,6 +89,40 @@ pub fn slice_nd_array(arr: ArrayD<Fr>, indices: &[usize]) -> ArrayD<Fr> {
   arr.slice_move(slice_info)
 }
 
+#[cfg(feature = "gpu")]
```
liliatangxy, line 91: I think this file is getting quite disorganized. We can split the gpu/cpu ops into a different file or figure out how to organize the functions based on their functionality

src/basic_block/cq.rs
```diff
@@ -16,12 +16,13 @@ use std::collections::HashMap;
 
 #[derive(Debug)]
 pub struct CQBasicBlock {
-  pub setup: Option<(i32, usize)>,
+  pub setup: ArrayD<Fr>,
```
liliatangxy: What is the reason for this change?

src/basic_block/div.rs
```diff
@@ -11,22 +12,39 @@ pub struct DivScalarBasicBlock {
...
+        }
+        div.push(Fr::from(z));
+        rem.push(Fr::from(r));
       }
-      div.push(Fr::from(z));
-      rem.push(Fr::from(r));
     }
+    #[cfg(feature = "gpu")]
+    let (div, rem): (Vec<_>, Vec<_>) = inputs[0]
+      .into_par_iter()
```
liliatangxy: Is there a reason why we can't have another wrapper for `iter` and `map`?

src/util.rs
Vimalakirti: This is a good point. I am working on it. But I think it would be better if we create another PR to handle this issue considering the current PR is already big.

src/basic_block/cq.rs
Vimalakirti: Fixed. (This change is for checking `RangeBasicBlock` in another ongoing branch. It was accidentally introduced in this PR.)

src/util.rs
Vimalakirti: The issue is put here:
https://github.com/uiuc-kang-lab/zk-torch/issues/52

src/basic_block/cq.rs
```diff
@@ -39,27 +39,32 @@ impl BasicBlock for CQBasicBlock {
...
+    util::ssm_g1_in_place(&mut Q_i_x_1, &scalars);
     let mut L_i_x_1 = srs.X1P[..N].to_vec();
     util::ifft_in_place(domain_N, &mut L_i_x_1);
     let mut L_i_0_x_1 = L_i_x_1.clone();
+    let scalars = (0..N).into_par_iter().map(|i| domain_N.group_gen_inv().pow(&[i as u64])).collect();
+    util::ssm_g1_in_place(&mut L_i_0_x_1, &scalars);
+
     let temp = srs.X1P[N - 1] * Fr::from(N as u64).inverse().unwrap();
-    L_i_0_x_1.par_iter_mut().enumerate().for_each(|(i, x)| *x = *x * domain_N.group_gen_inv().pow(&[i as u64]) - temp);
+    L_i_0_x_1.par_iter_mut().enumerate().for_each(|(_i, x)| *x -= temp);
```
liliatangxy: can remove the `i` variable name here and in cq2

src/util.rs
```diff
@@ -119,28 +160,77 @@ pub fn fft_helper<G: ScalarMul + std::ops::MulAssign<Fr>>(a: &mut Vec<G>, domain
...
+  if inv {
+    let gpu_result = G::gpu_ifft(domain, &a);
+    *a = gpu_result;
+  } else {
+    let gpu_result = G::gpu_fft(domain, &a);
+    *a = gpu_result;
+  }
+}
+
+#[cfg(not(feature = "gpu"))]
```
liliatangxy: i think we should be consistent with putting the gpu version before the not gpu version of each function

liliatangxy (approved):

src/util.rs
Vimalakirti: Sounds good, it will also be addressed in the next PR related to https://github.com/uiuc-kang-lab/zk-torch/issues/52

src/basic_block/matmul.rs
```diff
@@ -6,8 +6,44 @@ use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
 use ark_ff::Field;
 use ark_poly::{univariate::DensePolynomial, DenseUVPolynomial, EvaluationDomain, GeneralEvaluationDomain};
 use ark_std::{ops::Mul, ops::Sub, UniformRand, Zero};
-use ndarray::{ArrayD, Ix1, Ix2, IxDyn};
+use ndarray::{arr1, arr2, ArrayD, Ix1, Ix2, IxDyn};
 use rand::{rngs::StdRng, SeedableRng};
+#[cfg(feature = "gpu")]
+use rayon::prelude::*;
+
+#[cfg(not(feature = "gpu"))]
```
ddkang: For my understanding, why are these different?

src/basic_block/repeater.rs
```diff
@@ -2,8 +2,217 @@ use super::{BasicBlock, Data, DataEnc, PairingCheck, ProveVerifyCache, SRS};
...
 use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
 use ark_poly::univariate::DensePolynomial;
-use ndarray::{arr1, azip, s, ArrayD, Axis, Dimension, IxDyn, SliceInfo, SliceInfoElem};
+#[cfg(feature = "gpu")]
+use itertools::multiunzip;
+use ndarray::{arr1, azip, par_azip, s, ArrayD, Axis, Dimension, IxDyn, SliceInfo, SliceInfoElem};
 use rand::rngs::StdRng;
+use rayon::prelude::*;
+
+#[cfg(not(feature = "gpu"))]
```
ddkang: Ditto, why are these different

ddkang: This entire file

src/basic_block/matmul.rs
Vimalakirti: Moved all (feature = "gpu") tags to util

src/basic_block/repeater.rs
Vimalakirti: Moved all (feature = "gpu") tags to util

src/basic_block.rs
```diff
@@ -75,7 +78,8 @@ pub enum CacheValues {
   Data(Data),
   G2(G2Affine),
 }
-pub type ProveVerifyCache = HashMap<String, CacheValues>;
+
+pub type ProveVerifyCache = Arc<Mutex<HashMap<String, CacheValues>>>;
```
ddkang, line 86: Please document why this needs to be in an arc/mutex

liliatangxy:
I think it looks good minus Daniel's comment about arc/mutex. How was the gpu feature tested?

Vimalakirti:
> I think it looks good minus Daniel's comment about arc/mutex. How was the gpu feature tested?

It had been tested on our campus cluster gpus.

src/basic_block.rs
Vimalakirti: Documented

liliatangxy:
With what workloads?

Vimalakirti:
> With what workloads?

Had tested it with two sub-networks in RNNT (predictor and joint). Let me prepare some stats for you.

ddkang:
> > With what workloads?
> 
> Had tested it with two sub-networks in RNNT (predictor and joint). Let me prepare some stats for you.

Can we set up a github workflow to test on a GPU machine?

Vimalakirti:
> > > With what workloads?
> > 
> > 
> > Had tested it with two sub-networks in RNNT (predictor and joint). Let me prepare some stats for you.
> 
> Can we set up a github workflow to test on a GPU machine?

Yes, we can. But the standard GitHub-hosted runners have no GPU. The organization admin can either add [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/adding-self-hosted-runners) or we can configure [larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/managing-larger-runners) on GitHub. The [per-minute rate (USD)](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) on GPU-powered Linux 4-core machines is $0.07. According to our current GitHub action run time, It would take about $0.25 for each `push`.

ddkang:
> > > > With what workloads?
> > > 
> > > 
> > > Had tested it with two sub-networks in RNNT (predictor and joint). Let me prepare some stats for you.
> > 
> > 
> > Can we set up a github workflow to test on a GPU machine?
> 
> Yes, we can. But the standard GitHub-hosted runners have no GPU. The organization admin can either add [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/adding-self-hosted-runners) or we can configure [larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/managing-larger-runners) on GitHub. The [per-minute rate (USD)](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) on GPU-powered Linux 4-core machines is $0.07. According to our current GitHub action run time, It would take about $0.25 for each `push`.

Can we get this to run on our cluster?

Vimalakirti:
> Can we get this to run on our cluster?

Technically, we should be able to do this. GitHub suggests using Kubernetes and Helm for [managing our cluster runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/quickstart-for-actions-runner-controller). I've sent an email to campus cluster support to understand the possibility. Otherwise, the best resource I found is https://classee.cloud/resources/slurm/ , we may try setting it up later.

ddkang:
> > Can we get this to run on our cluster?
> 
> Technically, we should be able to do this. GitHub suggests using Kubernetes and Helm for [managing our cluster runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/quickstart-for-actions-runner-controller). I've sent an email to campus cluster support to understand the possibility. Otherwise, the best resource I found is https://classee.cloud/resources/slurm/ , we may try setting it up later.

Great, please let me know how this goes.

src/basic_block.rs
ddkang: Why is this shared between the prover and verifier? This is a highly broken design

src/basic_block/cq.rs
```diff
@@ -73,32 +78,36 @@ impl BasicBlock for CQBasicBlock {
...
     let L_i_0_x_1 = &setup.0[2 * N..];
-    let CacheValues::CQTableDict(table_dict) =
-      cache.entry(format!("cq_table_dict_{:p}", self)).or_insert_with(|| CacheValues::CQTableDict(HashMap::new()))
-    else {
-      panic!("Cache type error")
-    };
-    if table_dict.len() == 0 {
-      for i in 0..N {
-        table_dict.insert(model.raw[i], i);
+    let m_i = {
```
ddkang, line 81: Is this actually different or just a reformat? It's much harder to review when you have a bunch of reformatting. If this is a reformat, please do all the reformatting first so we can merge that first

src/basic_block.rs
Vimalakirti: Thanks for the feedback. I apologize for any confusion caused by the comments. I'd like to clarify that the cache is not shared between the prover and verifier. Instead, each maintains its own cache, which is shared only between threads within the same role (prover or verifier). This ensures efficient and safe access to the cache during concurrent operations.

To prevent any misunderstandings, I have updated the comments in the code to make this distinction clear.

Vimalakirti:
After #61, the PR introduces gpu functionalities in `util.rs`. All other files changed are due to the arc mutex wrapper for the ProveVerifyCache. The codes in util may seem overwhelming now, but this issue will be addressed in #59.

ddkang:
Looks good minus the pre-merge tests. Any updates from campus cluster on this?

Vimalakirti:
> Looks good minus the pre-merge tests. Any updates from campus cluster on this?

I sent the email on Friday morning, but they haven't responded yet. I will let you know once they do.

Vimalakirti:
> Looks good minus the pre-merge tests. Any updates from campus cluster on this?

They said Kubernetes is not available on CC and they don't think it will be added to the system. However, they mentioned seeing Apptainer being used to access SLURM commands from inside on other systems. 
As a follow-up, I asked if it's possible to run a server on CC to listen to requests from GitHub Actions (the solution in https://classee.cloud/resources/slurm/). Then, we could add our self-hosted runner on CC. I'll let you know once they respond.

Vimalakirti:
> Looks good minus the pre-merge tests. Any updates from campus cluster on this?

They said they will talk to the system administrators and see what is possible last night.

ddkang (approved):

ddkang:
If the code is ready, this is okay to merge. Let's figure out how to use AWS for testing.

Vimalakirti:
> If the code is ready, this is okay to merge. Let's figure out how to use AWS for testing.

Sounds good, the issue is created here:
https://github.com/uiuc-kang-lab/zk-torch/issues/71

PR #53, opened 2024-07-07
Maxpool

liliatangxy:
Added maxpool layer which implements and proves the maxpool reshape operations and the max operation itself.


src/basic_block/max.rs
```diff
@@ -18,3 +33,140 @@ impl BasicBlock for MaxBasicBlock {
...
+pub struct MaxProofBasicBlock;
+
+// This max includes a proof and is intended to be followed by a lookup range check on the second output.
+// TODO: Proof should be updated to use openings
+impl BasicBlock for MaxProofBasicBlock {
+  // Returns the max of the input and max - x for all x in input
+  fn run(&self, _model: &ArrayD<Fr>, inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
+    assert!(inputs.len() == 1 && inputs[0].ndim() == 1);
+    let max_arr = inputs[0]
+      .fold_axis(Axis(0), Fr::zero(), |max, y| if *y < Fr::from(1 << 28) && *y > *max { *y } else { *max })
```
Vimalakirti: what if elements in inputs[0] are all negative (e.g., [-1,-2,-3])? In this case, max should return -1 instead of zero.

src/basic_block/max.rs
```diff
@@ -18,3 +33,140 @@ impl BasicBlock for MaxBasicBlock {
...
+      .into_shape(vec![1])
+      .unwrap();
+    let max_val = max_arr.first().unwrap();
+
+    let mut r = ArrayD::zeros(inputs[0].shape());
+    azip!((r in &mut r, &x in inputs[0]) *r = *max_val - x);
+    vec![max_arr, r]
+  }
+
+  fn prove(
```
Vimalakirti: This seems complicated. Is there any reason why we cannot just impelment `run` for max and use SubBasicBlock and CQ protocol to ensure (max(original)-original)>=0 ?

src/basic_block/max.rs
liliatangxy: seems like the original `run` for `MaxBasicBlock` also has this issue. I'm not sure what we should consider to be the minimum element

src/basic_block/max.rs
liliatangxy: The problem with this approach is that we have to show that the product of the (max(original)-original) differences is 0 to show that the max is one of the values. We could implement a basic block that performs this behavior to modularize better or keep it as part of a max proof. I can imagine that argmax would also need that logic

liliatangxy: Realized that this proof was from before there was `RepeaterBasicBlock` and didn't adapt the product check correctly. Fixing that and it should make the code less complicated

src/basic_block/max.rs
liliatangxy: I fixed this, but I realized that the padding for maxpool is 0 and copy constraint expects padded values to be 0. This may cause issues when last axis is padded and the rest of the elements in the poly are negative

src/layer/pool.rs
```diff
@@ -0,0 +1,147 @@
...
+use crate::util::pad;
+use ark_bn254::Fr;
+use ark_std::Zero;
+use copy_constraint::zero_padding_partition;
+use ndarray::Axis;
+use ndarray::{arr1, indices, ArrayD, Dim, Dimension, IxDyn};
+use std::collections::BTreeMap;
+use tract_onnx::pb::AttributeProto;
+
+fn splat_input(input_shape: &Vec<usize>, strides: &Vec<usize>, pads: &Vec<usize>, ci: usize, ch_dims: &Vec<usize>) -> Vec<Vec<Option<IxDyn>>> {
```
Vimalakirti: Please add high-level descriptions for this function. Thx!

src/layer/pool.rs
```diff
@@ -0,0 +1,147 @@
...
+  if zero_indices.len() > 0 {
+    partitions.insert(Fr::zero(), zero_indices);
+  }
+  if nonzero_indices.len() > 0 {
+    partitions.insert(nonzero_val, nonzero_indices);
+  }
+  partitions
+}
+
+pub struct MaxPoolLayer;
```
Vimalakirti, line 82: Did you test it with an ONNX file?

liliatangxy: Yeah, lmk if you want me to send it to you

Vimalakirti: It will be great for me to prepare #62. Please send me all the layers you implemented and tested, thx!

Vimalakirti (dismissed):

src/layer/pool.rs
```diff
@@ -0,0 +1,147 @@
...
+use ark_std::Zero;
+use copy_constraint::zero_padding_partition;
+use ndarray::Axis;
+use ndarray::{arr1, indices, ArrayD, Dim, Dimension, IxDyn};
+use std::collections::BTreeMap;
+use tract_onnx::pb::AttributeProto;
+
+// This constructs the permutation for CopyConstraintBasicBlock to be inputted into MaxProofBasicBlock. The output is a (product of output dims of the pool operation * input channels X product of kernel_dims) permutation where the rows correspond to one max operation, and each row contains the set of arguments to max.
+// ci is the number of input channels
+fn splat_input(input_shape: &Vec<usize>, strides: &Vec<usize>, pads: &Vec<usize>, ci: usize, kernel_dims: &Vec<usize>) -> Vec<Vec<Option<IxDyn>>> {
```
ddkang, line 17: Is this used elsewhere, like in avg pool?

liliatangxy: Yeah, the same splat would be used for avg pool. I was planning on implementing it in the same file if we needed to implement it.

ddkang:
There are merge conflicts

ddkang (approved):

PR #59, opened 2024-07-17
Categorize util functions 

Vimalakirti:
Addressed the issue https://github.com/uiuc-kang-lab/zk-torch/issues/52. It should be merged after PR https://github.com/uiuc-kang-lab/zk-torch/pull/51.


src/graph.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
Vimalakirti, line 1: add this line to prevent the unused import warning for `verify_for_each_pairing`

ddkang (commented):
High level comments in the files please

ddkang:
Also we should have GitHub actions for tests

src/util/fft.rs
```diff
@@ -0,0 +1,223 @@
...
+ * The functions are used for performing FFT and IFFT on G1 and G2 points.
+ * Each function has a CPU and GPU implementation.
+ */
+#![allow(dead_code)]
+#![allow(unused_imports)]
+use ark_bn254::Fr;
+use ark_ec::ScalarMul;
+use ark_poly::{EvaluationDomain, GeneralEvaluationDomain};
+use rayon::prelude::*;
+#[cfg(feature = "gpu")]
```
ddkang, line 12: Is there a way to do an if else with the feature flags?

Vimalakirti: Yeah, there is, it looks like
```
if cfg!(feature = "gpu") {  do_something() } else { do_something_else() }
```
But I prefer not using this way here; otherwise, we will need to compile lots of icicle crates no matter the gpu feature is on or off.

ddkang (approved):

PR #60, opened 2024-07-20
Add Clip, Tan, and Identity

Vimalakirti:
Add Clip, Tan, and Identity and Handle special cases for Sub and Where layers
(network.onnx and challenge_18 are for testing these layers, they may be removed later)


src/basic_block/sub.rs
```diff
@@ -9,7 +9,9 @@ impl BasicBlock for SubBasicBlock {
   fn run(&self, _model: &ArrayD<Fr>, inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
     assert!(inputs.len() == 2 && inputs[0].ndim() <= 1 && inputs[1].ndim() <= 1);
     let mut r = ArrayD::zeros(IxDyn(&[std::cmp::max(inputs[0].len(), inputs[1].len())]));
-    if inputs[0].len() == 1 {
+    if inputs[0].len() == 1 && inputs[1].ndim() == 0 { // speicial case: [1] - []
```
ddkang: Comment in next line

src/layer/mul.rs
```diff
@@ -36,7 +36,17 @@ impl Layer for MulLayer {
       }),
       N: 1,
     }));
-    let mul_output = graph.addNode(if input_shapes[1].len() == 0 { mul_scalar } else { mul }, vec![(-1, 0), (-2, 0)]);
+    let mul_basicblock = if input_shapes[1].len() == 0 || input_shapes[0].len() == 0 {
```
ddkang, line 40: What's going on here?

src/main.rs
```diff
@@ -110,11 +110,20 @@ fn verify(srs: &SRS, graph: &Graph) {
 }
 
 fn main() {
-  let srs = &ptau::load_file("challenge", 7, 7);
-  let onnx_file_name = "sample.onnx";
+  let srs = &ptau::load_file("challenge_18", 18, 18);
```
ddkang: What's going no here?

Vimalakirti: I manually generate fake inputs for the mobius `network.onnx` because its inputs cannot be randomly generated. For example, it requires `maximum_response_time` > `response_time` > `minimum_response_time`. If the inputs do not follow this, it would cause an assertion error in our basicblocks. Besides, it requires a  larger challenge file because we need to multiply a large constant somewhere.

But this is just for testing, when merging this PR, I will revert `main.rs` to convert the original `sample.onnx`. Then, `challenge_18` and `network.onnx` will be removed.

src/layer/mul.rs
Vimalakirti: Our original code did not consider the case that inputs[0] might also be dim-0 value (a scalar). This modification considers this possibility. Added comments.

src/main.rs
ddkang: Why don't we make something modular that tests multiple ONNX files that cover the range of layers?

Vimalakirti: Yeah, I think we really need that. Lilia and I discussed this issue the other day. I will create a new PR for this feature soon.

Vimalakirti: The ongoing PR: https://github.com/uiuc-kang-lab/zk-torch/pull/62

Vimalakirti:
This PR will be closed because we should split it into smaller PRs.

PR #62, opened 2024-07-22
Add testing for each ONNX op

Vimalakirti:
Besides what's stated in the title, this PR adds the following new features
- changed the name of this repo from zkml_proofs to zk-torch in Cargo.toml
- fixed some shape bugs in lstm and concat (for dim not equal to 2^n)
- Added lib.rs file and moved all tests to `tests/`. `test_basicblocks` is the original `tests.rs`.

Currently, it's a draft, I'll add more ONNX operations into `tests/ops/` for testing to make sure every layer we supported is tested.


tests/test_basicblocks.rs
```diff
@@ -0,0 +1,230 @@
+#[cfg(test)]
```
ddkang, line 1: Shouldn't we be able to test this by comparing against the output of running tract or something?

Vimalakirti: yeah, it helps check the correctness of an ONNX op output.

ddkang:
Please fix the bugs in a different PR

PR #65, opened 2024-07-24
Add einsum

Vimalakirti:
The PR added einsum layer, which supports vector outer product (what's in GPTj) and vector inner product for now.


src/layer/einsum.rs
```diff
@@ -0,0 +1,163 @@
...
+use crate::graph::*;
+use crate::layer::{squeeze::UnsqueezeBasicBlock, Layer};
+use crate::onnx;
+use crate::util;
+use ark_bn254::Fr;
+use ndarray::ArrayD;
+use std::collections::HashMap;
+use tract_onnx::pb::AttributeProto;
+
+fn parse_einsum(equation: &str) -> (Vec<String>, Vec<String>) {
```
ddkang, line 11: Is there a function that does this in tract or otherwise?

src/layer/einsum.rs
```diff
@@ -0,0 +1,163 @@
...
+pub struct EinsumLayer;
+impl Layer for EinsumLayer {
+  fn graph(input_shapes: &Vec<&Vec<usize>>, _constants: &Vec<Option<&ArrayD<Fr>>>, attributes: &Vec<&AttributeProto>) -> (Graph, Vec<Vec<usize>>) {
+    let mut graph = Graph::new();
+    let equation = &attributes.iter().filter(|x| x.name == "equation").next().unwrap().s;
+    let equation = std::str::from_utf8(&equation).unwrap();
+    let (input_eqs, output_eq) = parse_einsum(equation);
+    assert!(input_eqs.len() == input_shapes.len());
+    // vector outer product
+    if input_eqs == vec!["a", "b"] && output_eq == vec!["ab"] {
```
ddkang, line 159: This seems complicated. Why don't you make functions for each?

src/layer/einsum.rs
Vimalakirti: I am not sure if their implementation suits our needs, but I will take a look at it.

src/layer/einsum.rs
```diff
@@ -0,0 +1,172 @@
...
+      assert!(input_shapes[0].len() == 1 && input_shapes[1].len() == 1);
+      let output_shape = vector_outer_product(&mut graph, input_shapes);
+      (graph, vec![output_shape])
+    // vector inner product
+    } else if input_eqs == vec!["a", "a"] {
```
liliatangxy: Do you need to check that the output is empty?

Vimalakirti: thx, we do need that. Added it.

ddkang (approved):

PR #68, opened 2024-07-24
Proof for max

liliatangxy:
Added MaxProofBasicBlock which enables the max operation with a proof. Its structure is similar to the Plonk/copy constraint proof.


src/basic_block/max.rs
```diff
@@ -18,3 +29,241 @@ impl BasicBlock for MaxBasicBlock {
     .into_dyn()]
   }
 }
+
+#[derive(Debug)]
+pub struct MaxProofBasicBlock;
+
+// This max includes a proof and is intended to be followed by a lookup range check on the second output.
+impl BasicBlock for MaxProofBasicBlock {
+  // Returns the max of the input and max - x for all x in input
```
Vimalakirti, line 38: For my understandings, why do we need max - x here? Could we compute it during proving?

src/basic_block/max.rs
```diff
@@ -18,3 +29,241 @@ impl BasicBlock for MaxBasicBlock {
...
+  fn run(&self, _model: &ArrayD<Fr>, inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
+    assert!(inputs.len() == 1 && inputs[0].ndim() == 1);
```
Vimalakirti, line 40: I am thinking how to find the max in a 2D array.
If inputs[0].ndim() should be 1, shall we first use copy constraint to paste 2D elements into a 1D array, and then use this block?

src/basic_block/max.rs
```diff
@@ -18,3 +29,241 @@ impl BasicBlock for MaxBasicBlock {
...
+    vec![Z_x].serialize_uncompressed(&mut bytes).unwrap();
+    proof.push(Z_x);
+    util::add_randomness(rng, bytes);
+
+    let alpha = Fr::rand(rng);
+
+    // t constraints:
+    // Z(oX) (s(X) + gamma) = (d(X) + gamma) * Z(X)
+    // L0(X)(Z(X)-1) = 0s
+    // L0(X)s(X) = 0
```
Vimalakirti, line 159: Is 0s a typo? And does the o in Z(oX) mean omega?

src/basic_block/max.rs
```diff
@@ -18,3 +29,241 @@ impl BasicBlock for MaxBasicBlock {
...
+    util::add_randomness(rng, bytes);
+    let zeta = Fr::rand(rng);
+
+    // Round 5 randomness
+    let mut bytes = Vec::new();
+    proof.2.serialize_uncompressed(&mut bytes).unwrap();
+    util::add_randomness(rng, bytes);
+    let v = Fr::rand(rng);
+
+    // Round 5 end randomness
```
Vimalakirti, line 271: Why are there two Round 5?

src/basic_block/max.rs
```diff
@@ -18,3 +29,241 @@ impl BasicBlock for MaxBasicBlock {
...
+    let mut bytes = Vec::new();
+    vec![W_x, W_gx].serialize_uncompressed(&mut bytes).unwrap();
+    util::add_randomness(rng, bytes);
+    let u = Fr::rand(rng);
+
+    // Verify that diff = max - input
```
Vimalakirti, line 277: I guess we can do assertion here (like what we do in SubBasicBlock).

Vimalakirti (commented):
Added some initial comments, will take a closer look at the protocol later,

src/basic_block/max.rs
liliatangxy: It's so that we can pass it to a CQBasicBlock to check that it is non-negative

src/basic_block/max.rs
liliatangxy: There are not two round 5s. The u value is generated at the end of round 5 because only the verifier uses it.

Vimalakirti: I see, if the prover does not use the random values, could we skip round 5 during proving

liliatangxy: The verifier needs the commitments constructed in Round 5 (W_z, W_gz). The random value u is used by the verifier because it batches checking these commitments with an addition in the pairing check. This technique is also used on pg. 31 step 12 of the Plonk paper.

Vimalakirti: (I meant we might delete line 190~193)

liliatangxy: Oh, those are needed so that in subsequent basic blocks, the rng state is the same between the prover and verifier.

src/basic_block/max.rs
liliatangxy: I think an alternative is proving max for each row, and then CopyConstraint the resulting max column into a 1D array and proving the max of that. I'm not sure which is faster though.

src/basic_block/max.rs
ddkang: If it's confusing, add it as a comment

src/basic_block/max.rs
ddkang: If it's confusing, add it as a comment

src/basic_block/max.rs
Vimalakirti: It seems complicated if we want to find a max over an N-dimensional array especially when N is really large

Vimalakirti (approved):
This PR introduces a way to prove max for a 1-d array F.
Let me summarize the high-level ideas here:
1. IFFT the 1-d array F as a polynomial f(X)
2. Compute another polynomial d(X) by IFFTing max - F and require
  - every evaluation of d(X) over the domain elements >= 0 --> this requires a range check such as CQ
  - proving the subtraction is correctly performed by [d]_1 == [max]_1 - [f]_1 ([]_1 denotes MSM w/ G1 elements)
3. Show max really exists in one of the evaluations of f(X) by showing 0 exists in d(X)
  - sort d(x) in ascending order to get s(x), then the first element in s(x) should be 0. --> need an opening
  - need to prove d(x) is a permutation of s(x). --> need openings
4. Prove all openings are valid

It looks good to me, but I guess it's really important to make sure people use this BB with additional CQBB for the range check mentioned in step 2. Otherwise, it is not secure.

ddkang:
Let's put these in the comments. What does "use this BB with additional CQBB for the range check mentioned in step 2" mean?

liliatangxy:
We need to range check the differences max - x for each input with cq to check that it is nonnegative when it is used in a layer. This is already mentioned in the comment above run.

ddkang:
I mean put the overall algorithm at the top, since this is a difficult proof:
IFFT the 1-d array F as a polynomial f(X)
Compute another polynomial d(X) by IFFTing max - F and require
every evaluation of d(X) over the domain elements >= 0 --> this requires a range check such as CQ
proving the subtraction is correctly performed by [d]_1 == [max]_1 - [f]_1 ([]_1 denotes MSM w/ G1 elements)
Show max really exists in one of the evaluations of f(X) by showing 0 exists in d(X)
sort d(x) in ascending order to get s(x), then the first element in s(x) should be 0. --> need an opening
need to prove d(x) is a permutation of s(x). --> need openings
Prove all openings are valid

liliatangxy:
Added a comment for the overview of the proof based on that summary

ddkang (approved):

PR #69, opened 2024-07-24
Add range proving

Vimalakirti:
Addressed the issue in https://github.com/uiuc-kang-lab/zk-torch/issues/58


Vimalakirti:
Let me briefly explain the proving idea here. 
First, we view the range outputs as evaluations of f(x) over domain generator omega^n. And let selection(x)=1 for all x=omega^i where i in [0, # of elements in the range minus 1]; selection(x)=0, otherwise.
Then the high-level idea is to prove selection(x) * [f(x)+step-f(omega * x)] will always be zero for all x=omega^n over the whole domain.
Let F(x) = selection(x) * [f(x)+step-f(omega * x)] , we can turn the above problem into proving a poly t(x) exists and F(x)=t(x)V(x), where V(x) is the vanishing polynomial over the domain. To do this, the proving procedure can be divided into 2 steps:
1. Given a random z sampled by verifier, we open F, t, and V at z and check F(z)==t(z)V(z).
2. We prove the openings in step 1 are all valid.

Vimalakirti:
Added RangeConstBasicBlock, which allows us to skip all proving when all start, limit, and delta are known during the graph compilation.

src/basic_block/range.rs
```diff
@@ -1,26 +1,232 @@
...
+
+// RangeConstBasicBlock is a basic block that creates a tensor of a range of values.
+// The range is defined by three constants: the start, limit, and delta values.
+#[derive(Debug)]
+pub struct RangeConstBasicBlock {
+  pub start: i32,
+  pub limit: i32,
+  pub delta: i32,
+}
+impl BasicBlock for RangeConstBasicBlock {
```
ddkang, line 37: How does this work for the verifier? Does the verifier need to do the MSM to prove this?

ddkang: Can this be done in the setup?

Vimalakirti: Added in setup

Vimalakirti:
I just realized that we can't use this RangeBasicBlock in our current implementation of zk-torch because we can't compile the graph without knowing the limit value. Otherwise, we can't infer the output shape during model compilation. But fortunately, we won't see an unknown limit in MLPerf models during model compilation. I added comments on top of RangeBasicBlock and removed it from RangeLayer.

ddkang:
> I just realized that we can't use this RangeBasicBlock in our current implementation of zk-torch because we can't compile the graph without knowing the limit value. Otherwise, we can't infer the output shape during model compilation. But fortunately, we won't see an unknown limit in MLPerf models during model compilation. I added comments on top of RangeBasicBlock and removed it from RangeLayer.

Can we make it so it errors out if we can't support it?

Vimalakirti:
> Can we make it so it errors out if we can't support it?

Yeah, we currently use RangeConstBasicBlock when all three fields are computable during graph compilation. And RangeLayer will panic when any of these fields are non-fixed.

src/basic_block/range.rs
```diff
@@ -1,26 +1,289 @@
...
+// The difference between RangeBasicBlock and RangeConstBasicBlock is that RangeBasicBlock
+// takes the limit value as a private input, while RangeConstBasicBlock takes the limit value as a constant.
+// We don't use this basic block in the current implementation because we can't compile the graph without knowing the limit value.
+// TODO: add proper blinding for opening arguments, similar issue as in CopyConstraintBasicBlock
+#[derive(Debug)]
+pub struct RangeBasicBlock {
+  pub start: i32,
+  pub delta: i32,
+}
+impl BasicBlock for RangeBasicBlock {
```
ddkang: Do we need this if everything is always constant for our ONNX layers? I'd prefer to omit this for now and keep it in a branch if we can stick with constants

ddkang:
Looks good to me, please have lilia take a look

Vimalakirti:
> Looks good to me, please have lilia take a look

@liliatangxy hi, lilia, please take a look at this pr, thx!

liliatangxy (approved):

ddkang (approved):

PR #72, opened 2024-07-26
Add Xor layer

Vimalakirti:

PR #73, opened 2024-07-26
Add And layer

Vimalakirti:


src/onnx.rs
```diff
@@ -125,6 +125,7 @@ fn get_local_graph(
 ) -> (Graph, Vec<Vec<usize>>) {
   match op {
     "Add" => Ok(AddLayer::graph(&input_shapes, &node_constants, &node_attributes)),
+    "And" => Ok(MulLayer::graph(&input_shapes, &node_constants, &node_attributes)), // use MulLayer for And
```
ddkang: Ditto: Does this require that the inputs are in {0, 1}? Is this implicitly implied by the inputs?

Vimalakirti: Probably the easiest way is we use a basicblock to perform the boolean check during model compilation.

src/basic_block/bool_check.rs
```diff
@@ -0,0 +1,20 @@
+use super::{BasicBlock, Data, DataEnc, PairingCheck, ProveVerifyCache, SRS};
```
ddkang: Is this secure? If not, please make an issue and note it in a comment

Vimalakirti: I think this depends on how we define security. We say a layer is secure if:
(1) It correctly performs the defined functionalities, or
(2) It correctly performs the defined functionalities and checks the correctness of inputs.

According to definition (1), the previous layer should be responsible for the correctness of the inputs to the current layer. If all previous layers are secure, then the correctness of inputs is delegated, ensuring the inputs of the entire model are correct.

According to definition (2), each layer should verify its input format.

I added new `prove` and `verify` for Boolean checks considering our consensus might be (2).

Vimalakirti: Added an issue #83 because we do not need bool check(s) after a Cast layer.

liliatangxy:
lgtm minus the blinding comment in https://github.com/uiuc-kang-lab/zk-torch/pull/74

Vimalakirti:
> lgtm minus the blinding comment in #74

Thanks, comments added.

ddkang:
Now that we know the fix, why can't we fix it now?

Vimalakirti:
Fixed the opening issue. Now it is secure.

Vimalakirti:
From now on, please ignore #72 and #74. Let us focus on And and get it merged first.

liliatangxy (dismissed):
Blinding looks good

ddkang (approved):

PR #74, opened 2024-07-26
Add Not layer

Vimalakirti:


src/layer/not.rs
```diff
@@ -0,0 +1,27 @@
+use crate::basic_block::*;
```
ddkang, line 1: Ditto: Does this require that the inputs are in {0, 1}? Is this implicitly implied by the inputs?

Vimalakirti: Probably the easiest way is we use a basicblock to perform the boolean check during model compilation.

src/basic_block/bool_check.rs
```diff
@@ -0,0 +1,20 @@
+use super::{BasicBlock, Data, DataEnc, PairingCheck, ProveVerifyCache, SRS};
+use crate::util;
+use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
+use ndarray::{arr0, azip, ArrayD, IxDyn};
+use rand::rngs::StdRng;
+
+// BooleanCheckBasicBlock is a basic block that checks if all elements are 0 or 1
```
ddkang: See comment on bool check in the other PR

Vimalakirti: see the response in #73

src/basic_block/bool_check.rs
```diff
@@ -0,0 +1,141 @@
...
+    // Fiat-Shamir
+    let mut bytes = Vec::new();
+    let mut proof_1 = vec![t_x + srs.Y1P * r[0]];
+    proof_1.serialize_uncompressed(&mut bytes).unwrap();
+    proof_g1.append(&mut proof_1);
+    util::add_randomness(rng, bytes);
+
+    // open f and t at zeta, so that the verifier can check that t(zeta) * zeta^N - 1 = f(zeta) * (1 - f(zeta))
+    let zeta = Fr::rand(rng);
+    let f_poly_zeta = inputs[0].first().unwrap().poly.evaluate(&zeta);
```
liliatangxy: Does f need to be blinded? If so, we need to add a similar comment about the security that copy constraint has

Vimalakirti: See response in #73

liliatangxy (dismissed):
good once we use And's bool check BB

ddkang (approved):

PR #76, opened 2024-07-26
Add Flatten layer

Vimalakirti:


src/layer/flatten.rs
```diff
@@ -0,0 +1,49 @@
...
+  let permutation = ArrayD::from_shape_fn(input_shape, |index| Some(index));
+  let permutation = permutation.view().into_shape(&output_shape.clone()[..]).unwrap().to_owned();
+  let padded_permutation = util::pad_to_pow_of_two(&permutation, &None);
+
+  (padded_permutation, output_shape)
+}
+
+// https://onnx.ai/onnx/operators/onnx__Flatten.html
+// Flattens the input tensor into a 2D matrix.
+// If input tensor has shape (d_0, d_1, … d_n) then the output will have shape (d_0 X d_1 … d_(axis-1), d_axis X d_(axis+1) … X dn).
```
ddkang: I don't think this parses

liliatangxy: I think we can add 'X's before and after the '...'s

liliatangxy (approved):
lgtm after fixing the comment

src/layer/flatten.rs
Vimalakirti: it was rewritten

ddkang (approved):

PR #80, opened 2024-07-27
Handle a special-case shape issue for SubBB

Vimalakirti:
Our `SubBasicBlock` accept two inputs. When input[0].len()=1, we assumed that input[1].len() will not be 1. This PR included the case that we didn't consider before.


ddkang (commented):
Can we test this?

Vimalakirti:
> Can we test this?

we can test this with an ONNX that matches the description above

liliatangxy:
Why can't we test this with a `testBasicBlock` test?

Vimalakirti:
Thanks for the reviews, a new test case was added.

liliatangxy (approved):

ddkang (approved):

PR #81, opened 2024-07-27
Handle special-case shape issues using Mul

Vimalakirti:
ONNX Mul accepts two inputs `a` and `b`, and returns `a*b`. It performs broadcasting for inputs when the shapes don't match.
We used MulBasicBlock when `a` and `b` are both 1-dimensional arrays and used MulScalarBasicBlock when `b` is a scalar. However, we didn't consider the case that `a` might also be a scalar in Mul and Where layers. 
This PR addressed these issues.

### DIFF
--- a/src/layer/mul.rs
+++ b/src/layer/mul.rs
@@ -36,7 +36,20 @@ impl Layer for MulLayer {
       }),
       N: 1,
     }));
-    let mul_output = graph.addNode(if input_shapes[1].len() == 0 { mul_scalar } else { mul }, vec![(-1, 0), (-2, 0)]);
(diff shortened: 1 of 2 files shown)


ddkang (commented):
Can we test this?

Vimalakirti:
> Can we test this?

we can test this with an ONNX that matches the description above

liliatangxy:
Why can't we test this with a `testBasicBlock` test?

Vimalakirti:
> Why can't we test this with a `testBasicBlock` test?

The files changed are all under `layer/`

liliatangxy (approved):

ddkang (approved):

PR #82, opened 2024-07-28
Support arbitrary padding values in copy constraint

liliatangxy:
Previously, copy constraint could only support pad values of 0. However, operations such as max and min may produce incorrect outputs when they perform operations containing both actual values of the tensor and pad values of 0.

Also fixes a bug introduced in the LSTM PR with a cq block in Conv layer


src/basic_block/copy_constraint.rs
```diff
@@ -63,7 +63,7 @@ fn construct_ssig(
   idxs: &[((usize, usize), Option<(usize, usize)>)],
   N: usize,
   last_dim: usize,
-  partitions: &HashMap<Option<(usize, usize)>, Vec<(usize, usize)>>,
+  partitions: &HashMap<Option<(usize, usize)>, Vec<Vec<(usize, usize)>>>,
```
ddkang, line 68: This needs some comments

ddkang (commented):
I think this needs some tests

src/basic_block/copy_constraint.rs
```diff
@@ -90,23 +89,52 @@ fn construct_ssig(
     .collect()
 }
 
+// Returns the padding_partitions field for CopyConstraintBasicBlock when the given permutation padding elements are 0
+pub fn zero_padding_partition(permutation: &ArrayD<Option<IxDyn>>) -> BTreeMap<Fr, Vec<IxDyn>> {
```
Vimalakirti, line 92: Is there any reason that we don't put this in `util/`?

src/layer/conv.rs
```diff
@@ -147,7 +152,7 @@ impl Layer for ConvLayer {
     }));
     let change_SF_check = graph.addBB(Box::new(RepeaterBasicBlock {
       basic_block: Box::new(CQBasicBlock {
-        setup: Array1::from_iter(-(1 << 10)..1 << 11).map(|x| Fr::from(*x)),
+        setup: Array1::from_iter(onnx::CQ_RANGE_LOWER..-onnx::CQ_RANGE_LOWER).map(|x| Fr::from(*x)),
```
Vimalakirti, line 155: Is there any reason why we want this array to be in descending order?

liliatangxy: That should be ascending order since `CQ_RANGE_LOWER` is the min

src/basic_block/copy_constraint.rs
liliatangxy: Wasn't sure whether it makes more sense to put it in util or here given that it is only ever used with CCBB. I can move it to util if that is more appropriate

liliatangxy:
> I think this needs some tests

Currently there is a test that uses the `zero_padding_partition` function which creates one padding value, another test for a permutation with two different padding values, and tests with no padding values. Not sure what other tests to add

Vimalakirti:
When we have no need to use index, we might skip `.enumerate()` after `iter`. Lgtm minus this.

Vimalakirti (approved):

Vimalakirti (approved):

ddkang (approved):

Issue #71, opened 2024-07-25
GPU testing


Vimalakirti:
To test our gpu feature, we should integrate aws gpu runners into our GitHub Actions.

ddkang:
We can also see if we can get GPUs on GitHub

PR #84, opened 2024-07-29
Test gpu by enqueueing jobs on slurm

Vimalakirti:
This PR is to address #71


Vimalakirti:
> This PR is to address #71

But our CC will not echo back the running results to GitHub Actions. We need to manually check if the gpu test is successful. 
To check it:
- Install [ntfy.sh](https://ntfy.sh/) on your phone.
- Subscribe topic "zk_torch_gpu"
- ntfy will pop out "GPU test starts on SLURM" if the test begins to run on a SLURM node
- ntfy will pop out "GPU test completes" if the test ends without being killed
- Please ssh to CC and cd to ddkang/ under /projects/.../
- Then cd to bjchen4/gh_actions/{COMMIT-HASH}/ to see the log `zktorch_gh_action.out`

.github/workflows/gpu.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs: 
+  enqueue:
+    runs-on: ubuntu-latest
+    steps: 
+      - uses: actions/checkout@v2
+      - name: Get a short version of the GIT commit SHA to use in naming files
+        id: getshortsha
+        run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+      - name: Copy files over to the cluster
+        uses: garygrossgarten/github-action-scp@v0.6.0
```
ddkang: What are these specific ones?

ddkang:
nfty streams are public so this an easy way to get your phone spammed. Also why can't we have the success fail in the notification?

Vimalakirti:
> nfty streams are public so this an easy way to get your phone spammed. Also why can't we have the success fail in the notification?

Thanks for the comment. Yeah, I need to search for another proper way to send notifications. And we can send the results in the notification if we parse the log file. Let me fix it.

.github/workflows/gpu.yml
Vimalakirti: Let me explain each step here:
1. `uses: actions/checkout@v2`: this pulls the repo to the VM hosted by GitHub
2. `name: Get a short version of the GIT commit SHA`: this gets a 8 chars commit hash to rename our repo later so that we don't mess up different commits on CC when testing
3. `name: Copy files over to the cluster`: this `scp` the repo on the VM to CC
4. `name: Execute script to enqueue job`: this `ssh` to CC. And it appends necessary dependencies for gpu testing and sends the sbatch job to CC SLURM node to test it

Vimalakirti:
This PR is now ready for reviews. There are good and bad news for this new feature.
- good news: it does not cost a penny to run gpu tests and we can check whether a test passed in our slack channel (`zk-torch-test-gpu` in `mldeployment`). I'll add you to the channel later (and feel free to mute this channel because it might be annoyed when lots of commits were tested).
- bad news: the network of campus cluster is really not stable. Sometimes, GitHub Actions may fail to connect to CC for random reasons. The only way is to re-run the gpu test a few minutes later.
For more info, please check the `README.md` under `test_gpu_scripts/`

ddkang (commented):
Does GitHub actions run the CPU tests? Or should we add those here too for the layers and such

Vimalakirti:
> Does GitHub actions run the CPU tests? Or should we add those here too for the layers and such

Yeah, just added `cargo test --features gpu`, it'll run all the necessary tests. 
Now we have two GitHub action files under `.github/workflows/`. 
- `rust.yml` will execute `cargo run` and `cargo test` on GitHub VM for cpu testing.
- `gpu.yml` will `scp` the repo and enqueue `cargo run` and `cargo test` to Campus Cluster for gpu testing.

ddkang (approved):

PR #85, opened 2024-07-31
Max layer

liliatangxy:
Implements max layer in the case where there are exactly two inputs and the second input is a constant of size 1. This is the only form it shows up as in RetinaNet


src/layer/max.rs
```diff
@@ -0,0 +1,97 @@
...
+      let cc = graph.addBB(Box::new(CopyConstraintBasicBlock {
+        permutation,
+        input_dim: IxDyn(&input_shape_padded),
+        padding_partitions,
+      }));
+      let max = graph.addBB(Box::new(RepeaterBasicBlock {
+        basic_block: Box::new(MaxProofBasicBlock {}),
+        N: 1,
+      }));
+      let reshape_shape = &vec![input_shapes[0].iter().product(), 1];
```
Vimalakirti, line 39: What happened if input_shapes[0] is vec![3, 3, 3]? I guess the two ccbb compilation would fail in this case.

liliatangxy: Why does this fail? The reshape takes into account the actual input and output shapes before adding the padding

Vimalakirti: Yeah, I re-checked the entire splat fn and the reshape, and found I thought wrong.

Vimalakirti (dismissed):

src/layer/max.rs
```diff
@@ -0,0 +1,97 @@
...
+use crate::layer::Layer;
+use crate::util::pad_to_pow_of_two;
+use ark_bn254::Fr;
+use ark_std::Zero;
+use copy_constraint::zero_padding_partition;
+use ndarray::{concatenate, indices, ArrayD, Axis, Dim, Dimension, IxDyn};
+use std::collections::BTreeMap;
+use tract_onnx::pb::AttributeProto;
+
+// Returns the splat needed to pass into MaxProofBasicBlock. This produces a (product of input dims X 2) permutation where the first column corresponds to the input elements and the second column contains None for the constant values
```
ddkang, line 10: Is this code repeated? If so, let's refactor it

liliatangxy: This one isn't but the other helper function is repeated so I created a new utils file for copy constraint helpers. I'll start to move shared ones into there.

ddkang (approved):

Vimalakirti (approved):

PR #87, opened 2024-08-04
Add TopK layer

Vimalakirti:


src/layer/topk.rs
```diff
@@ -0,0 +1,108 @@
...
+
+  let topk_idx = ArrayD::from_shape_fn(output_shape.as_slice(), |index| Some(index.clone()));
+  let padded_topk_idx = util::pad_to_pow_of_two(&topk_idx, &None);
+  assert!(padded_topk_idx.shape() == padded_output_shape.as_slice());
+
+  padded_topk_idx
+}
+
+// TopKLayer is a layer that returns the top k elements of the input tensor along a given axis
+// The order of the elements is determined by the 'largest' attribute, the default '1' means descending
```
liliatangxy, line 26: please give an overview of the proof

src/basic_block/one_to_one.rs
```diff
@@ -0,0 +1,366 @@
...
+
+    // compute g_idx_s(x) = f_idx_s(x) + r_f_idx_s(x) * (x^N - 1) for secure opening
+    let r_f_idx_s_poly = DensePolynomial {
+      coeffs: vec![r[9], r[10], r[11]],
+    };
+    let g_idx_s_poly = inputs[3].first().unwrap().poly.clone().add(r_f_idx_s_poly.mul_by_vanishing_poly(domain));
+    let g_idx_s_x = util::msm::<G1Projective>(&srs.X1A, &g_idx_s_poly.coeffs);
+
+    let minus_one_poly = DensePolynomial::from_coefficients_vec(vec![-Fr::one()]);
+    // compute q(x) = [f(x) - g(x)]/ (x^N - 1) for proving f(x) = g(x)
```
liliatangxy, line 87: maybe these equality checks can be batched with the other equality checks?

liliatangxy (commented):
Initial review, will finish by tonight

src/basic_block/one_to_one.rs
Vimalakirti: We discussed about this issue. Since we need to use them separately in the Z(x) proving, it may be more time consuming to batch them (these g polys) here.

ddkang: Please leave a comment

src/basic_block/ordered.rs
```diff
@@ -0,0 +1,266 @@
...
+      coeffs: vec![outputs[0].first().unwrap().r],
+    });
+    let r_diff_plus_r_Q_diff_x = util::msm::<G1Projective>(&srs.X1A, &r_diff_plus_r_Q_diff.coeffs);
+
+    // Compute g_s(omega * x) polynomial
+    let g_s_omega_poly = DensePolynomial {
+      coeffs: g_s_poly.coeffs.iter().enumerate().map(|(i, x)| x * &domain.element(i)).collect(),
+    };
+
+    // Calculate L0Z(x) = L0(x)(Z(x)-1) polynomial
```
liliatangxy: i think L0 should be L_{N-1} and there is no Z
why do we want L_{N-1}(x) = 1?

Vimalakirti: Thanks. Fixed the comment and script.

The reason we needed this was:
originally, when x=omega^(N-1), diff(x)=g_s(omega^(N-1)) - 0, which not equals to g_s(x) - g_s(omega * x).
But this comment inspired me to just compute diff(x) to be g_s(x) - g_s(omega*x) for all x over the domain in `run`, which saves us from the need of L_{N-1}.
The TL;DR is we don't need L_{N-1}(x) anymore.

src/basic_block/one_to_one.rs
Vimalakirti: Added

liliatangxy (dismissed):

ddkang (dismissed):

Vimalakirti:
The returned output shape in TopKLayer should not be padded. The latest commit fixed this issue.

liliatangxy (approved):

PR #88, opened 2024-08-07
Min layer

liliatangxy:
Implements min layer in the case where there are exactly two inputs and the second input has size 1 but isn't a constant. We additionally have to require that the first input has ndim > 1 since we utilize concat.

This is the only form it shows up as in RetinaNet, where the second input comes from a Shape layer followed by a Gather layer. It would be helpful if @Vimalakirti could double-check this logic.


src/layer/max.rs
```diff
@@ -56,3 +58,79 @@ impl Layer for MaxLayer {
...
+
+      let extended_second_input = ArrayD::from_shape_vec(input_shape_padded.clone(), vec![Some(IxDyn(&[0])); size]).unwrap();
+
+      let extend_second_input = graph.addBB(Box::new(CopyConstraintBasicBlock {
+        permutation: extended_second_input,
+        input_dim: IxDyn(&[1]),
+        padding_partitions: BTreeMap::new(),
+      }));
+
+      let concat_inputs = graph.addBB(Box::new(ConcatBasicBlock { axis: 0 }));
```
Vimalakirti, line 84: If input[0] and input[1] are both length 1 arr1, then we can't use concatBB directly.

liliatangxy: The condition in L67 checks for this. I believe this also will not occur in RetinaNet since the left input will always be the max output which I don't think is arr1.

ddkang: I think we should comment this.

Vimalakirti: I thought wrong because I didn't consider L67. Thanks for your explanation.

Vimalakirti (dismissed):
This PR looks good. My only concern is MinLayer takes a little more proving time than MaxLayer does because we need two Neg to transform this into a Max problem. But the extra time seems to be ignorable.

src/layer/max.rs
ddkang: Please comment and it's good to go

liliatangxy: Ready

ddkang (approved):

Vimalakirti (approved):

PR #89, opened 2024-08-07
Add Mod layer

Vimalakirti:


src/layer/modulo.rs
```diff
@@ -0,0 +1,88 @@
+use crate::basic_block::*;
```
ddkang: Is this code basically copied from div? If so can we have a unified thing that only returns one or the other?

Vimalakirti: Thanks for the comment, they've been unified by a macro.

liliatangxy (dismissed):
i think just leave a comment about output_idx

ddkang (approved):

PR #91, opened 2024-08-10
Faster CopyConstraint

Vimalakirti:
This PR makes CopyConstraintBB faster. It introduces two changes:

1. Leverage a tmp hashmap instead of iterating `self.padding_partitions` for every index. Improved `run` time a lot (O(n^2)->O(n)).
2. Replace BTreeMap by HashMap. Improved `prove` and `verify` time a little (O(logn)->O(1) for map.get(.)).


ddkang:
Is this deterministic?

liliatangxy:
`padding_vals` will make the key iteration order between the prover and verifier the same, which is sufficient for correct RLC batching in the protocol. I'm not sure if it matters that the keys can still appear in a different order between runs due to different vector orderings

ddkang:
It makes certain kinds of testing harder but maybe that's fine

ddkang:
How much of a performance improvement is the BTree vs HashMap?

Vimalakirti:
> How much of a performance improvement is the BTree vs HashMap?

I would say almost no improvement for our current layers because our Map uses only few `Fr` currently (like only 1 or 2). If it's not necessary, I can revert HashMap to BTreeMap.

ddkang:
If it makes it deterministic I think it's good

Vimalakirti:
Yeah, I think it's deterministic based on our current implementation. I'm planning to further sort the map keys in prove and verify. This saves us from the extra field `padding_vals` and guarantees the order when iterating the map by default.

ddkang (approved):

liliatangxy (approved):

PR #92, opened 2024-08-11
Add Gemm layer

Vimalakirti:


src/layer/gemm.rs
```diff
@@ -0,0 +1,133 @@
...
+use tract_onnx::pb::AttributeProto;
+
+// Gemm computes Y = alpha * A' * B' + beta * C, where
+// the first input tensor A has shape (M, K) or (K, M),
+// the second input tensor B has shape (K, N) or (N, K),
+// (optional) the third input tensor C is broadcastable to shape (M, N),
+// and output tensor Y has shape (M, N).
+// A will be transposed to A' before doing the computation if attribute transA is non-zero, same for B and transB.
+pub struct GemmLayer;
+impl Layer for GemmLayer {
```
ddkang, line 17: This doesn't do a general Gemm does it? If not, we should be explicit with what we allow and don't

Vimalakirti: Thanks for noticing! But this does realize the general Gemm (as described [here](https://onnx.ai/onnx/operators/onnx__Gemm.html)).

liliatangxy (approved):

ddkang (approved):

PR #93, opened 2024-08-12
Add InstanceNorm layer

Vimalakirti:


src/layer/norm.rs
```diff
@@ -213,3 +215,271 @@ impl Layer for BatchNormLayer {
     (graph, vec![input_shapes[0].clone()])
   }
 }
+
+// InstanceNorm is a struct that represents an instance normalization layer, which computes
+// Y = (X - mean) * scale / sqrt(var + epsilon) + bias,
+// where mean and var are computed per instance per channel
+pub struct InstanceNormLayer;
+impl Layer for InstanceNormLayer {
```
liliatangxy, line 223: Is there a lot of repeated code between InstanceNorm and BatchNorm?

Vimalakirti: The main differences are the calculation for mean, variance, and the division per instance per channel (BatchNorm is per channel). But I agree there are some repeated codes such as addition of epsilon and multiplication with scale. Do you think it's better to introduce a macro for them?

liliatangxy: No, I think it's different enough then

ddkang (approved):
lgtm if @liliatangxy approves

liliatangxy (approved):

PR #94, opened 2024-08-12
Print stats and separate setup and run from prove

Vimalakirti:
as the title stated

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,7 +1,7 @@
 #![allow(non_snake_case)]
 #![allow(non_upper_case_globals)]
 use crate::graph::Graph;
-use ark_bn254::{Fr, G1Affine, G2Affine};
+use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
 use ark_poly::univariate::DensePolynomial;
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
 use basic_block::*;
@@ -11,7 +11,7 @@ use rayon::prelude::*;
 use sha3::{Digest, Keccak256};
 use std::fs::{self, File};
 use std::io::Read;
-use util::convert_to_data;
+use util::{convert_to_data, measure_file_size};
 mod basic_block;
 mod graph;
 mod layer;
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -116,7 +152,16 @@ fn main() {
...
+  setup(&srs, &graph, &models);
+  let outputs = run(&inputs, &graph, &models);
+  prove(&srs, &inputs, outputs, &mut graph);
   verify(&srs, &graph);
+  // measure_proof_size
+  let _modelsEncSize = measure_file_size("modelsEnc");
+  let inputsEncSize = measure_file_size("inputsEnc");
+  let outputsEncSize = measure_file_size("outputsEnc");
+  let proofsSize = measure_file_size("proofs");
+  let totalSize = inputsEncSize + outputsEncSize + proofsSize;
```
Vimalakirti: I think the proof size only includes the size of inputsEnc, outputsEnc, and proofs. But I may be wrong.

ddkang: @liliatangxy can you check this

src/main.rs
```diff
@@ -71,7 +101,11 @@ fn prove(srs: &SRS, inputs: &Vec<&ArrayD<Fr>>, graph: &mut Graph, models: &Vec<&
   let mut rng = StdRng::from_seed(buf);
 
   // Prove:
+  let start = std::time::Instant::now();
   let proofs = graph.prove(srs, &setups, &models, &inputs, &outputs, &mut rng);
+  let prove_time = start.elapsed();
+  stat_println!("Time taken to prove: {:?}", prove_time);
+  stat_println!("Total time taken to prove: {:?}", encode_time + prove_time);
```
Vimalakirti: I think the proving time should include the time to msm I/O and to run prove. But I may be wrong.

src/main.rs
```diff
@@ -21,16 +21,36 @@ mod ptau;
 mod tests;
 mod util;
 
-fn prove(srs: &SRS, inputs: &Vec<&ArrayD<Fr>>, graph: &mut Graph, models: &Vec<&ArrayD<Fr>>) {
-  // Run:
-  let outputs = graph.run(inputs, models);
+macro_rules! stat_println {
```
ddkang: Aren't there good libraries for this? I believe plonky2/3 have good examples

ddkang (commented):
I think we should use a library since the library can determine if it should print based on env variables, etc

src/main.rs
ddkang: What do you mean by MSM i/o? If we assume steady state, we don't need to time the time to load the SRS

Vimalakirti: It does not include the time to load the SRS. It means we commit the input tensor, intermediate tensors, and the output tensor by msm them.
(I/O is ambiguous, I should say msm input and output tensors)

src/main.rs
Vimalakirti: I just checked Plonky2, they add a feature tag above their print function (i.e., `#[cfg(feature = "timing")]`). Do you think this simple fix could address this issue?

ddkang: Don't they also have a way to preprint the timing?

src/main.rs
ddkang: Ah okay, yes we should include that but not for the one-time setup.

src/main.rs
Vimalakirti: What do you mean by "preprint"? Do you mean print all the timing results together?

ddkang: typo - pretty print

Vimalakirti: I checked Plonky2 and 3 again. I don't think they have pretty print. What they have above ours is the feature tag above print, and the use of debug print somewhere.

Vimalakirti: But maybe we can use [this](https://crates.io/crates/prettytable-rs) to print them in an organizing way.

src/main.rs
liliatangxy: I think it should just contain proofs

src/main.rs
liliatangxy: Plonky2 has `TimingTree` to print out proving time breakdowns: https://github.com/0xPolygonZero/plonky2/blob/main/plonky2/src/util/timing.rs#L8

Vimalakirti: Thanks! Let me check how to use this struct.

Vimalakirti:
Thanks for all the valuable comments. The PR is now ready for reviews.
The new changes:
- adopted `plonky2` `TimingTree`, which eases us to show the proving time and its breakdown
- fixed proof size to be only for `proofs`
- use nightly build for GitHub actions because `plonky2` needs it.

liliatangxy (approved):

ddkang (approved):

PR #98, opened 2024-08-14
Resize layer

liliatangxy:
Adds Resize layer for coordinate_transformation_mode = asymmetric, mode = nearest, and nearest_mode = floor (the only use case in stable diffusion)


src/basic_block/copy_constraint.rs
```diff
@@ -88,21 +88,6 @@ fn construct_ssig(
     .collect()
 }
 
-// Returns the padding_partitions field for CopyConstraintBasicBlock when the given permutation padding elements are 0
-pub fn zero_padding_partition(permutation: &ArrayD<Option<IxDyn>>) -> HashMap<Fr, Vec<IxDyn>> {
```
Vimalakirti: Is there any reason to move zero_padding_partition into util? (I remember we discussed this before.)

liliatangxy: Now that we have the copy_constraint utils file I think it makes sense to put it there since it is now used for many layers

Vimalakirti: It makes sense. But there's one concern: since some layers in other PRs (e.g., Argmax) also use this fn, I think it might cause bugs if we merge this after those PRs. Do you think separating this feature from this PR helps? Or is there any better way?

liliatangxy: I can separate it into a different PR

Vimalakirti (approved):

ddkang (approved):

PR #99, opened 2024-08-15
Print out layer names

Vimalakirti:
We didn't print layer names during running, proving, and verifying, which made it difficult to debug when running large models (we only printed the names of basicblocks). This PR implements this feature so that we can check it on netron when a layer fails

### DIFF
--- a/src/onnx.rs
+++ b/src/onnx.rs
@@ -115,6 +115,7 @@ fn create_output_indices<'a>(
       basic_block: graph.basic_blocks.len(),
       inputs: vec![],
     });
+    graph.layer_names.push(format!("Const {}", name.to_string()));
     graph.basic_blocks.push(Box::new(ConstBasicBlock {}));
   }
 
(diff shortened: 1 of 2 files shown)


src/onnx.rs
```diff
@@ -291,9 +293,10 @@ fn process_node(
 
   // handle a special case (op == "Shape")
   if op == "Shape" {
+    let shape = arr1(&input_shapes[0].iter().map(|&x| Fr::from(x as i32)).collect::<Vec<_>>()).into_dyn();
     passed_constants.insert(
       (&node.output[0]).to_string(),
-      arr1(&input_shapes[0].iter().map(|&x| Fr::from(x as i32)).collect::<Vec<_>>()).into_dyn(),
+      util::pad_to_pow_of_two(&shape, &Fr::zero()),
```
Vimalakirti, line 299: Besides, this PR also adds a shape padding for Shape layer here

liliatangxy (dismissed):

Vimalakirti:
also print layer name during model compilation

liliatangxy (approved):

PR #100, opened 2024-08-16
Convert copy-constraint padding partitions into an enum

liliatangxy:
There are currently only two modes of cc padding partitions used, so we can convert it into an enum to reduce memory

### DIFF
--- a/src/basic_block/copy_constraint.rs
+++ b/src/basic_block/copy_constraint.rs
@@ -22,6 +22,7 @@ use rayon::prelude::*;
 use std::{
   cmp::{max, min},
   collections::{BTreeMap, HashMap},
+  default,
   iter::{once, repeat, Map},
 };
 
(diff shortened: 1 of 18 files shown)


src/basic_block/copy_constraint.rs
```diff
@@ -103,16 +104,56 @@ pub fn zero_padding_partition(permutation: &ArrayD<Option<IxDyn>>) -> HashMap<Fr
...
+  Zero,
+  Max(Fr),
+}
+
+// Gets padding partition based on the enum value
+fn get_padding_partition(permutation: &ArrayD<Option<IxDyn>>, pad_type: &PaddingEnum) -> HashMap<Fr, Vec<IxDyn>> {
+  match pad_type {
+    PaddingEnum::Zero => zero_padding_partition(permutation),
+    PaddingEnum::Max(val) => max_padding_partitions(permutation, *val),
+  }
```
Vimalakirti, line 156: Is it deterministic for setup, run, prove, and verify?

liliatangxy: For setup and run, the hashmap order is not relevant. For prove and verify, the fix you made previously with sorting the keys should still work.

Vimalakirti: How about the order of `Vec<IxDyn>`?

liliatangxy: I think the ndarray iterators are deterministic. They satisfy row-major order: https://docs.rs/ndarray/latest/ndarray/struct.ArrayBase.html#indexing-and-dimension

Vimalakirti (approved):

ddkang (approved):

PR #101, opened 2024-08-17
Fix shape bugs

Vimalakirti:
Fixed a lot of shape issues across the repo. (mainly due to padding and special cases)

### DIFF
--- a/src/onnx.rs
+++ b/src/onnx.rs
@@ -75,7 +75,9 @@ fn parse_onnx_constants<'a>(
       DatumType::F32 => {
         let tensor = tensor.into_array::<f32>().unwrap();
         Ok(tensor.map(|x| {
+          // handle the case where the constant is very close to zero (i.e., epsilon to prevent division by zero)
           if *x < 1e-10 && *x > 0.0 {
+            // the reason we use 1 here is because it is the smallest positive value that can be represented in the field
             return Fr::from(1);
           }
           let mut y = (*x * SF_FLOAT).round();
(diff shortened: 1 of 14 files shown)


src/onnx.rs
```diff
@@ -75,6 +75,9 @@ fn parse_onnx_constants<'a>(
       DatumType::F32 => {
         let tensor = tensor.into_array::<f32>().unwrap();
         Ok(tensor.map(|x| {
+          if *x < 1e-10 && *x > 0.0 {
```
ddkang, line 79: What'as this?

src/util/shape.rs
```diff
@@ -3,6 +3,7 @@
  * The functions are used for shape-related operations, such as
  * slicing and padding arrays.
  */
+#![allow(dead_code)]
```
ddkang: Why this?

src/onnx.rs
Vimalakirti: When the scale factor is not big enough, an epsilon constant will be 0, which might cause DivisionByZero error. This is a workaround solution for Bert and GPTj.

src/util/shape.rs
Vimalakirti: We currently do not use the function `slice_nd_array`,  but we might need to use it later. `#![allow(dead_code)]` is a tag which helps ignore the warning that a function is unused.

src/onnx.rs
ddkang: This minimally needs a comment as to why this is 1

src/util/shape.rs
ddkang: Can you put it above the relevant function?

liliatangxy (dismissed):

src/onnx.rs
Vimalakirti: added comments

src/util/shape.rs
Vimalakirti: moved the tag above the function

ddkang (approved):

PR #102, opened 2024-08-17
Save cq/cq2 setups for nonlinearities and load config

Vimalakirti:
This PR added a feature that saves CQ/CQ2 setups, which prevents us from wasting time for setting up the same nonlinear layers across different models.

### DIFF
--- a/src/basic_block/cq.rs
+++ b/src/basic_block/cq.rs
@@ -1,11 +1,11 @@
 #![allow(non_snake_case)]
 #![allow(non_upper_case_globals)]
 use super::{BasicBlock, CacheValues, Data, DataEnc, PairingCheck, ProveVerifyCache, SRS};
+use crate::onnx;
 use crate::util;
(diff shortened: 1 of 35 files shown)


src/basic_block/cq.rs
```diff
@@ -25,6 +27,15 @@ impl BasicBlock for CQBasicBlock {
   }
 
   fn setup(&self, srs: &SRS, model: &ArrayD<Data>) -> (Vec<G1Projective>, Vec<G2Projective>, Vec<DensePolynomial<Fr>>) {
+    let file_name = format!("{}.setup", util::hash_str(&format!("{self:?}")));
```
ddkang: I dislike how all of the file loading is hard coded. We should have a config file that handles this (which would also allow for multiple different models/scale factors). Please fix this now or make an issue

Vimalakirti: Let me fix this issue now.

liliatangxy (approved):

ddkang (approved):

PR #103, opened 2024-08-19
CopyConstraint setup optimizations

liliatangxy:
- remove extraneous Lagrange basis computations


Vimalakirti:
> * remove extraneous Lagrange basis computations

How fast would it be after this?

PR #104, opened 2024-08-23
Add DatumType to Layer constants

liliatangxy:
When we process constants in the onnx compiler, we scale floats based on the scale factor but not ints when converting them to their field values. This causes issues for layers such as Div which involve potential scale factor changes and can accept either float or int arguments.

The solution here is to pass the onnx DatumType to `graph()` for the `constants` parameter for layer-specific logic.

Also added the Div int fix


Vimalakirti:
I think it's ok to fix the mentioned Div issue in this PR.

Vimalakirti (approved):
lgtm

ddkang (approved):

PR #106, opened 2024-08-27
Precompute sigma map

Vimalakirti:
While working on copy constraint BB, I noticed some potential areas for performance improvement in the function `construct_ssig`. I made some modifications that resulted in a significant reduction in setup time (a 99.8% decrease in runtime during local testing on ResNet18). Specifically, this can help reduce thousands of seconds per layer when the copy constraint is really big.

Here’s a summary of the changes:

- I optimized the function by precomputing the sigma hashmap for the function `construct_ssig`.
This replaced the current O(n^2) algorithm with a more efficient one O(n) for computing the hashmap.
I believe these changes could help zk-torch run more efficiently and benefit the entire team. However, I understand that there may be other considerations I might have missed, and I would really appreciate @liliatangxy 's feedback on this.
- Besides, I also modified some `iter()` into `par_iter()` for further acceleration

Thanks for Lilia's great work—it was well-organized and easy to understand, which made it straightforward to identify these optimization opportunities.

### DIFF
--- a/src/basic_block/copy_constraint.rs
+++ b/src/basic_block/copy_constraint.rs
@@ -66,19 +66,15 @@ fn construct_ssig(
   idxs: &[((usize, usize), Option<(usize, usize)>)],
   N: usize,
   last_dim: usize,
-  partitions: &HashMap<Option<(usize, usize)>, Vec<Vec<(usize, usize)>>>,
+  sigma_map: &HashMap<Option<(usize, usize)>, HashMap<(usize, usize), (usize, usize)>>,
   is_input: bool,
 ) -> Vec<Fr> {
   let domain = GeneralEvaluationDomain::<Fr>::new(N).unwrap();
   idxs
-    .iter()
+    .par_iter()
     .flat_map(|(idx, perm_idx)| {
       let inp_idx = if is_input { Some(*idx) } else { *perm_idx };
-      let sigma = {
-        let partition = partitions.get(&inp_idx).ok_or_else(|| format!("Key {:?} not found in the partition", inp_idx)).unwrap();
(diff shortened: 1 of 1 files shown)


liliatangxy (dismissed):

liliatangxy:
I think you can also merge #103 with these changes, though I don't think #103 has much performance improvement

src/basic_block/copy_constraint.rs
```diff
@@ -240,9 +236,21 @@ impl BasicBlock for CopyConstraintBasicBlock {
     Zip::from(&mut inp_arr).and(&inp_idxs).for_each(|r, &a| {
       *r = (a, None);
     });
+    let sigma_map: HashMap<Option<(usize, usize)>, HashMap<(usize, usize), (usize, usize)>> = partitions
```
ddkang, line 238: Leave a comment about why we're doing this here and efficiency. Otherwise LGTM

src/basic_block/copy_constraint.rs
```diff
@@ -66,19 +66,15 @@ fn construct_ssig(
   idxs: &[((usize, usize), Option<(usize, usize)>)],
   N: usize,
   last_dim: usize,
-  partitions: &HashMap<Option<(usize, usize)>, Vec<Vec<(usize, usize)>>>,
+  sigma_map: &HashMap<Option<(usize, usize)>, HashMap<(usize, usize), (usize, usize)>>,
```
liliatangxy, line 67: need to update the above comment

Vimalakirti:
Added comments and integrated the performance improvement (removal of unnecessary ffts and msms by @liliatangxy ) from #103

liliatangxy (approved):

ddkang (approved):

PR #107, opened 2024-08-28
Pass input types to layers and fix Mul layer scaling based on type

liliatangxy:
Arithmetic layers like Mul assume that inputs are floats scaled by the scale factor, so when ints are passed to these layers (e.g., output of Shape layer) they will be incorrectly scaled down.

This PR
- implements passing input types between layers
- only adds scaling down to the graph in Mul if its inputs are floats

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -88,7 +88,8 @@ pub mod xor;
 pub trait Layer {
   fn graph(
     input_shapes: &Vec<&Vec<usize>>,
+    input_types: &Vec<DatumType>,
     constants: &Vec<Option<(&ArrayD<Fr>, DatumType)>>,
     attributes: &Vec<&AttributeProto>,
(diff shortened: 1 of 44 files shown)


Vimalakirti:
Do we still need DatumType in layer Constant? Could we get the same information from input_types?

liliatangxy:
> Do we still need DatumType in layer Constant? Could we get the same information from input_types?

In Range layer the layer's output type depends on the constant type

Vimalakirti (dismissed):

ddkang:
> > Do we still need DatumType in layer Constant? Could we get the same information from input_types?
> 
> In Range layer the layer's output type depends on the constant type

Let's document this in the code then

ddkang (approved):

Vimalakirti (approved):

PR #108, opened 2024-08-29
Stable Diffusion vae_decoder fixes

liliatangxy:
Additional changes needed to run vae_decoder:
- Support some ops when ndim = 0
- Support when end shape contains 0 for Reshape and support when startshape len > endshape len
- Support using constants for some layers in place of attributes or inputs

Also some additional cleanup missed previously


src/layer/reshape.rs
```diff
@@ -13,12 +14,48 @@ impl Layer for ReshapeLayer {
...
     input_types: &Vec<DatumType>,
     constants: &Vec<Option<(&ArrayD<Fr>, DatumType)>>,
-    _attributes: &Vec<&AttributeProto>,
+    attributes: &Vec<&AttributeProto>,
   ) -> (Graph, Vec<Vec<usize>>, Vec<DatumType>) {
     let mut graph = Graph::new();
 
     let startShape = input_shapes[0];
-    let mut endShape: Vec<_> = constants[1].unwrap().0.as_slice().unwrap().iter().map(|x| util::fr_to_int(*x)).filter(|x| *x != 0).collect();
+    let allowzero = match attributes.iter().filter(|x| x.name == "allowzero").next() {
```
Vimalakirti: I feel what `allowzero` does is not straightforward. Could you add some comments or link the doc?

Vimalakirti (commented):
Left a comment. Otherwise lgtm.

src/layer/reshape.rs
liliatangxy: I will simplify the code since allowzero is optional and only ever used with its default value in the model

ddkang (approved):

Vimalakirti (approved):

PR #109, opened 2024-08-29
Skip precomputable op and gen fake weight

Vimalakirti:
This PR contains new features and fixes for GPTj, which includes:
- skipping setup, proving, and verify for precomputable basicblocks and layers
- shape issues (support broadcastable Pow, general Expand, and get the correct val attribute for ConstOfShape)
- generation of fake weights when the model weights are from external files

But there's still one issue: we still encode (msm) the precomputable data. 
This issue is a little bit complicated: we cannot directly skip it for precomputable layers as the `encodeOutputs` of some basicblocks require encoded inputs from previous layers such as `concat`. If we skip it, some `encodeOutputs` will just fail because of no accessible encoded inputs.

Add this PR now because SDXL also requires fake weight generation and some fixes for shape issues may also help. I'll fix the above issue in another PR after I complete the works on copy constraint BB.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -24,7 +24,7 @@ pub struct Node {
 #[derive(Debug)]
 pub struct Graph {
   pub basic_blocks: Vec<Box<dyn BasicBlock>>,
-  pub precomputed_bb: Vec<bool>,
+  pub precomputable: (Vec<bool>, Vec<bool>), // (for basic_blocks, for nodes)
   pub layer_names: Vec<String>,
   pub nodes: Vec<Node>,
   pub outputs: Vec<(i32, usize)>,
@@ -92,7 +92,7 @@ impl Graph {
       .zip(models.iter())
       .enumerate()
       .map(|(i, (b, m))| {
-        let precomputable = self.precomputed_bb[i];
+        let precomputable = self.precomputable.0[i];
         if precomputable {
           println!("skipping setup for {:?} {:?}", i, b);
           return (vec![], vec![], vec![]);
@@ -140,7 +140,7 @@ impl Graph {
       .iter()
       .enumerate()
       .map(|(i, n)| {
-        let precomputable = self.layer_names[i].contains("(precomputed)");
+        let precomputable = self.precomputable.1[i];
         if precomputable {
           println!("skipping proving for {:?} {:?}", i, self.basic_blocks[n.basic_block]);
           return (vec![], vec![], vec![]);
@@ -200,7 +200,7 @@ impl Graph {
       .iter()
       .enumerate()
       .map(|(i, n)| {
-        let precomputable = self.layer_names[i].contains("(precomputed)");
+        let precomputable = self.precomputable.1[i];
         if precomputable {
           println!("skipping verifying for {:?} {:?}", i, self.basic_blocks[n.basic_block]);
           return vec![];
@@ -255,7 +255,7 @@ impl Graph {
           }
         })
         .collect();
-      let precomputable = self.layer_names[i].contains("(precomputed)");
+      let precomputable = self.precomputable.1[i];
       if !precomputable {
         let pairings = self.basic_blocks[n.basic_block].verify(srs, models[n.basic_block], &myInputs, outputs[i], proofs[i], rng, cache.clone());
         let mut bytes = Vec::new();
@@ -277,7 +277,7 @@ impl Graph {
   pub fn new() -> Self {
     Graph {
       basic_blocks: vec![],
-      precomputed_bb: vec![],
+      precomputable: (vec![], vec![]),
       layer_names: vec![],
       nodes: vec![],
       outputs: vec![],
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -17,17 +17,17 @@ use std::sync::{Arc, Mutex};
 
 #[derive(Debug)]
 pub struct Precomputable {
-  pub for_setup: Vec<bool>,
-  pub for_prove_and_verify: Vec<bool>,
-  pub for_encodeOutputs: Vec<bool>,
+  pub setup: Vec<bool>,
+  pub prove_and_verify: Vec<bool>,
+  pub encodeOutputs: Vec<bool>,
 }
 
 impl Precomputable {
   pub fn new() -> Self {
     Precomputable {
-      for_setup: vec![],
-      for_prove_and_verify: vec![],
-      for_encodeOutputs: vec![],
+      setup: vec![],
+      prove_and_verify: vec![],
+      encodeOutputs: vec![],
     }
   }
 }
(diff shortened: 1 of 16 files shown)


Vimalakirti:
The new commit added more datatypes into `util::datatype_to_datumtype`

src/layer/constantofshape.rs
```diff
@@ -19,7 +19,7 @@ impl Layer for ConstOfShapeLayer {
   ) -> (Graph, Vec<Vec<usize>>, Vec<DatumType>) {
     let mut graph = Graph::new();
 
-    let value = Fr::from(attributes.iter().filter(|x| x.name == "value").next().unwrap().i);
+    let value = Fr::from(attributes.iter().filter(|x| x.name == "value").next().unwrap().t.clone().unwrap().raw_data[0]);
```
liliatangxy: might need to change the layer output type to match the tensor type

src/graph.rs
```diff
@@ -91,6 +92,11 @@ impl Graph {
       .zip(models.iter())
       .enumerate()
       .map(|(i, (b, m))| {
+        let precomputable = self.precomputed_bb[i];
+        if precomputable {
+          println!("skipping setup for {:?} {:?}", i, b);
+          return (vec![], vec![], vec![]);
```
ddkang, line 133: Is this correct?

ddkang: How do the system know what to retrieve?

src/graph.rs
```diff
@@ -239,25 +255,30 @@ impl Graph {
...
-      util::add_randomness(rng, bytes);
-      pairings.iter().for_each(|p| {
-        assert!(p
-          .iter()
-          .fold(PairingOutput::<Bn<ark_bn254::Config>>::zero(), |acc, x| {
-            acc + Bn254::pairing(x.0, x.1)
-          })
-          .is_zero());
-      });
+      let precomputable = self.layer_names[i].contains("(precomputed)");
```
ddkang: This seems very fragile

src/onnx.rs
```diff
@@ -89,21 +91,29 @@ fn parse_onnx_constants<'a>(
...
-        let tensor = tensor.into_array::<bool>().unwrap();
-        Ok((tensor.map(|x| Fr::from(*x as i32)), DatumType::Bool))
-      }
-      _ => Err(format!("Unsupported constant type: {:?}", tensor.datum_type())),
-    }
-    .unwrap();
+      .unwrap();
+      (tensor, data_type)
+    } else {
+      // if the data_location is not None, we generate fake weights for now.
```
ddkang, line 109: Add an explicit "TODO" or "FIXME"

ddkang: Also we should have a compile flag or something that errors at runtime for insecure/invalid computation that we want to fix later

src/graph.rs
Vimalakirti: Maybe we could introduce a new field for this rather than reuse `layer_names`

src/graph.rs
Vimalakirti: Yeah, this is correct because the i-th element in `graph.precomputable.0` records whether to setup the i-th basicblock or not. And the decision is made during graph compilation in `onnx.rs`.

liliatangxy:
Can you make a separate PR for the shape issues? That part should be quick to review

src/onnx.rs
Vimalakirti: Added comments and WARNING at runtime

Vimalakirti:
The new commit skips encoding (msm) the precomputable data.

src/util/onnx.rs
```diff
@@ -39,29 +38,36 @@ pub fn generate_fake_inputs_for_onnx(filename: &str) -> Vec<ArrayD<Fr>> {
...
+  };
+  ArrayD::from_shape_vec(shape, input).unwrap()
+}
+
 // Converts ints for the DataType enum into DatumType
 // https://docs.rs/tract-onnx/latest/tract_onnx/pb/tensor_proto/enum.DataType.html
 pub fn datatype_to_datumtype(t: i32) -> DatumType {
   match t {
-    1 => DatumType::F32,
+    2 | 3 | 4 | 5 | 7 | 12 | 13 => DatumType::I64,
```
liliatangxy: i think we can put 6 back in here

liliatangxy (dismissed):

src/util/onnx.rs
Vimalakirti: Put it back

liliatangxy (dismissed):

Vimalakirti:
Here's some before-and-after stats for RNNT encoder on my mac:
- proving time: 111.84s -> 17.19s (-84.6%)
- verification time: 3.39s -> 1.01s (-70.2%)
- proof size: 2.53MB -> 243KB (-90.6%)

src/graph.rs
```diff
@@ -60,8 +61,17 @@ impl Graph {
     outputs: &Vec<&Vec<&ArrayD<Fr>>>,
     timing: &mut TimingTree,
   ) -> Vec<Vec<ArrayD<Data>>> {
+    assert!(self.nodes.len() == self.precomputable.2.len());
     let mut outputsEnc = vec![vec![]; self.nodes.len()];
     self.nodes.iter().enumerate().for_each(|(i, n)| {
+      let precomputable = self.precomputable.2[i];
+      if precomputable {
+        println!(
+          "{} | skipping encodingOutputs for {i} {:?}",
```
ddkang: it should say why it's being skipped

src/graph.rs
ddkang: see comment above on saying why it's being skipped

src/onnx.rs
ddkang: Where is the warning?

src/onnx.rs
```diff
@@ -130,6 +140,9 @@ fn create_output_indices<'a>(
     });
     graph.layer_names.push(format!("Const {}", name.to_string()));
     graph.basic_blocks.push(Box::new(ConstBasicBlock {}));
+    graph.precomputable.0.push(false);
```
ddkang: Why are we using indexes for this and not a named tuple or anything else?

Vimalakirti: Made it a named tuple now

src/graph.rs
Vimalakirti: Added comments

src/graph.rs
Vimalakirti: Now we use a struct called `Precomputable`

src/graph.rs
Vimalakirti: Added comments

src/graph.rs
ddkang: The print should explain why something is being skipped, otherwise the user has no idea.

Vimalakirti: thanks for the valuable comment, fixed this.

src/graph.rs
```diff
@@ -15,6 +15,23 @@ use std::collections::HashMap;
 use std::fs::File;
 use std::sync::{Arc, Mutex};
 
+#[derive(Debug)]
+pub struct Precomputable {
+  pub for_setup: Vec<bool>,
```
liliatangxy: i think we can remove the for_ for all of these

liliatangxy (dismissed):

src/graph.rs
Vimalakirti: Removed.

liliatangxy (approved):

src/main.rs
```diff
@@ -200,6 +200,7 @@ fn main() {
   let srs = &ptau::load_file(&CONFIG.ptau.ptau_path, CONFIG.ptau.pow_len_log, CONFIG.ptau.loaded_pow_len_log);
   let onnx_file_name = &CONFIG.onnx.model_path;
   let (mut graph, models) = onnx::load_file(onnx_file_name);
+  // TODO: read inputs from json file
```
ddkang, line 203: This should leave a warning at runtime

ddkang (approved):

src/main.rs
Vimalakirti: When calling `util::generate_fake_inputs_for_onnx()`, a warning will be shown. But we can make it clearer in the next PR. The issue was created here (https://github.com/uiuc-kang-lab/zk-torch/issues/116).

PR #110, opened 2024-08-29
GPTJ shape issues

Vimalakirti:
Moved the shape fixes from #109

### DIFF
--- a/src/layer/constantofshape.rs
+++ b/src/layer/constantofshape.rs
@@ -1,11 +1,13 @@
 use crate::basic_block::*;
 use crate::graph::*;
 use crate::layer::Layer;
+use crate::onnx;
 use crate::util;
 use ark_bn254::Fr;
 use ndarray::ArrayD;
 use tract_onnx::pb::AttributeProto;
 use tract_onnx::prelude::DatumType;
+use tract_onnx::pb::tensor_proto::DataType;
 
 // Generate a tensor with a given value (the value is in the ONNX attribute) and shape (the shape is in the input tensor)
 // reference: https://onnx.ai/onnx/operators/onnx__ConstantOfShape.html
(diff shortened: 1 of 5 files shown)


src/layer/constantofshape.rs
```diff
@@ -19,7 +19,7 @@ impl Layer for ConstOfShapeLayer {
   ) -> (Graph, Vec<Vec<usize>>, Vec<DatumType>) {
     let mut graph = Graph::new();
 
-    let value = Fr::from(attributes.iter().filter(|x| x.name == "value").next().unwrap().i);
+    let value = Fr::from(attributes.iter().filter(|x| x.name == "value").next().unwrap().t.clone().unwrap().raw_data[0]);
```
liliatangxy: will need to change the layer's output type to be based on the tensor's type

Vimalakirti: added, please check

liliatangxy (dismissed):

src/layer/pow.rs
```diff
@@ -16,7 +41,34 @@ impl Layer for PowLayer {
     _attributes: &Vec<&AttributeProto>,
   ) -> (Graph, Vec<Vec<usize>>, Vec<DatumType>) {
     let mut graph = Graph::new();
-    assert!(constants[1].unwrap().0.first().unwrap() == &Fr::from(2 * *onnx::SF as u32));
+
+    // if both constants[0] and constants[1] are Some
```
ddkang, line 54: Not clear why you're doing this

liliatangxy: can comment something like if both operands are constants then we can precompute/omit proof

Vimalakirti: Added comments to explain. (Also included a TODO tag in comments)

ddkang: Is this insecure as is? If so, add a runtime warning.

Vimalakirti: It is secure because both inputs are constants during graph compilation.

ddkang (dismissed):

Vimalakirti:
The new commit included the support for dim-0 Add and Sub. (This is also a fix I used for the GPTj job on CC, but I forgot to include it in this PR.)

Vimalakirti:
I want to double check if we need to consider datumType in the constant of Pow.
That's the reason why this PR is turned to a draft.

liliatangxy (approved):

ddkang (approved):

PR #112, opened 2024-08-30
Optimize copy constraint

liliatangxy:
* avoid intermediate interpolation in multiplication
* binary tree-ordered multiplication
* precompute ssig evals

The before-and-after stats for this part for the first CCBB of resnet18 on Bing-Jyue's mac:
17668.51s -> 61.97s

PR #113, opened 2024-08-30
CCBB optimization

Vimalakirti:
More optimization for #112

### DIFF
--- a/src/basic_block/copy_constraint.rs
+++ b/src/basic_block/copy_constraint.rs
@@ -348,12 +348,8 @@ impl BasicBlock for CopyConstraintBasicBlock {
     Z[0] = Fr::one();
     for j in 0..(N - 1) {
       let o = domain.element(j);
-      let num: Fr = (0..io_evals.len()).into_par_iter()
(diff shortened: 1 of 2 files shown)


Vimalakirti:
The PR includes the following optimization:
- Prevent any polynomial evaluation during the computation of Z poly (the technique here is to send the evaluation of ssig from setup and evaluate i/o polys in parallel beforehand)
- Use fft and ifft in mul_polys to make eval_over_domain faster

Will include the benchmark stats based on my local testing later.

liliatangxy (approved):

Vimalakirti:
Here's the speed-up stats (testing on the first CCBB of resnet18) for the above optimization:
- Z poly generation: 302.03s -> 248.46ms (reduce 99.91% time)
- `mul_polys`: 17668.51s -> 10207.93s (reduce 42.22% time)

However, the `mul_polys` part is still the main bottleneck of CCBB. 
I will take a closer look at it.

ddkang (approved):

PR #114, opened 2024-09-01
Accelerate poly multiplication

Vimalakirti:
When creating the t polynomial in copy constraint BB, we need to compute two products of N degree-D polynomials. The part was the main bottleneck when the given permutation is big. This PR further accelerated the polynomial multiplication by improving the time complexity from O(NDlogD) to O(DlogDlogN). (btw, this PR also used rayon to parallelize the updated algorithm)

The before-and-after stats for this part (the first CCBB of resnet18 on my mac): 
10207.93s -> 61.97s (-99.39%)

### DIFF
--- a/src/util/poly.rs
+++ b/src/util/poly.rs
@@ -1,5 +1,5 @@
 use ark_bn254::Fr;
-use ark_ff::{Field, Zero};
+use ark_ff::Zero;
 use ark_poly::{univariate::DensePolynomial, DenseUVPolynomial, EvaluationDomain, GeneralEvaluationDomain, Polynomial};
 use ark_std::One;
 use rayon::prelude::*;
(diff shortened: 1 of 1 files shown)


liliatangxy (approved):

src/util/poly.rs
```diff
@@ -11,22 +11,42 @@ where
...
-  }
+  let N: usize = polys.iter().map(|p| p.coeffs.len()).sum();
+  let domain = GeneralEvaluationDomain::<Fr>::new(N).unwrap();
+
+  let p_evals = polys.par_iter().map(|p| domain.fft(&p.coeffs)).collect::<Vec<_>>();
+  let p_evals = elementwise_product(&p_evals);
   DensePolynomial::from_coefficients_vec(domain.ifft(&p_evals))
 }
+
+pub fn mul_polys(polys: &Vec<DensePolynomial<Fr>>) -> DensePolynomial<Fr> {
```
ddkang, line 32: This seems fine for now. Is this the most efficient way to implement it? If not, we should leave a TODO

Vimalakirti: Added comments

ddkang (approved):

PR #117, opened 2024-09-03
Fixes for Unet

liliatangxy:
* add support for Gather accepting negative indices
* generalize Transpose layer

### DIFF
--- a/src/layer/gather.rs
+++ b/src/layer/gather.rs
@@ -13,13 +13,7 @@ impl BasicBlock for GatherBasicBlock {
   fn run(&self, _model: &ArrayD<Fr>, inputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Fr>> {
     let mut v = Vec::new();
     inputs[1].for_each(|x| {
-      let idx = {
-        if *x > Fr::from(inputs[0].len() as i32) {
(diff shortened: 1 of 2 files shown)


Vimalakirti:
What is the case the previous Transpose layer cannot support?

liliatangxy:
It only covered the cases where the last axis was in the last two positions

src/layer/gather.rs
```diff
@@ -34,9 +40,14 @@ impl Layer for GatherLayer {
     let mut graph = Graph::new();
     let mut indices_output = -2;
     if input_shapes[1].len() == 0 {
-      let indices = graph.addBB(Box::new(Const2BasicBlock {
-        c: constants[1].unwrap().0.clone(),
-      }));
+      let indices = constants[1].unwrap().0.mapv(|x| {
```
Vimalakirti: If inputs[1] must be a constant in Gather layer, I think we may prefer to handle the negative index in Gather layer instead of in Gather basicblock because that's what we typically do in other layers. Plus, it's good to keep a basicblock as simple as possible.

liliatangxy: removed the logic from bb and removed the input condition in layer

Vimalakirti (approved):

ddkang (approved):

PR #120, opened 2024-09-09
SF related features

Vimalakirti:
The PR added the following new features/fixes related to scale factor issues:
- Fixed `cast.rs` (we didn't prove when change sf)
- Returned `SF_LOG` instead of `SF` for `F32` in `util::datatype_to_datumtype`
- Added new implementation for Softmax layer (the previous method easily caused cq range error)
- Added `run()` for cq and cq2 to detect cq range error before proving. We can skip it by setting `enable_precheck` in `config.yaml` to be `false`
- Loaded real input data (in `.json` format)
- Implemented the binary search algorithm for finding the optimal SF under a given CQ range

(Many files were changed because of the introduction of `lib.rs` and `bin/`)


Vimalakirti:
The current PR is difficult to review, please wait until I split the PR into smaller ones. Thanks!

Vimalakirti:
Closed the PR since it had been split into: #121 #122 #124 #126

PR #121, opened 2024-09-11
Add lib

Vimalakirti:
To add `bin/` later under our repo, this PR added `lib.rs`, where we will export our mods. That's why lots of files were updated. But this PR changes no functionality in any module.
Besides, this PR also replaced the zkml_proofs in Cargo.tomal with zk_torch.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,43 @@
+#![allow(non_snake_case)]
+#![allow(non_upper_case_globals)]
+#![allow(unused_imports)]
+pub mod basic_block;
+pub mod graph;
+pub mod layer;
+pub mod onnx;
+pub mod ptau;
+#[cfg(test)]
+pub mod tests;
+pub mod util;
+
+use once_cell::sync::Lazy;
+use std::env;
+use std::fs::{self, File};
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+pub mod tests;
+pub mod util;
+
+use once_cell::sync::Lazy;
+use std::env;
+use std::fs::{self, File};
+use std::io::Read;
+use std::path::Path;
+
+pub static CONFIG_FILE: Lazy<String> = Lazy::new(|| {
```
ddkang, line 19: We don't necessarily need this now, but you should use an actual command line parsing tool, not this

ddkang: Raise an issue if you don't want to do this now

ddkang (approved):

liliatangxy (approved):

src/lib.rs
Vimalakirti: Sounds good. I will address this issue once we merge these PR in. 
The issue was created here https://github.com/uiuc-kang-lab/zk-torch/issues/123

PR #122, opened 2024-09-11
Fixed SF issues

Vimalakirti:
- Fixed `cast.rs` (we need to use CQ when change sf)
- Returned `SF_LOG` instead of `SF` for F32 in `util::datatype_to_datumtype` (SF is not a correct param for nonlinear)
- Added new implementation for `Softmax` layer (the previous method easily caused cq range error)

### DIFF
--- a/src/layer/softmax.rs
+++ b/src/layer/softmax.rs
@@ -81,6 +81,22 @@ impl Layer for SoftmaxLayer {
       N: 1,
     }));
 
+    // The proving idea is as follows
+    // 1. m = max(X): 
+    //    We first compute the maximum value of the input array. 
+    // 2. X - m: 
+    //    We subtract the maximum value from each element of the input array.
+    // 3. e^(X - m): 
+    //    We compute the exponential of each element of the input array. 
+    //    And we use "exp_check" to ensure that the output is within the CQ range.
+    // 4. SUM(e^(X - m)): 
+    //    We compute the sum of the exponential of each element of the input array.
+    // 5. 1 / SUM(e^(X - m)):
+    //    We compute the reciprocal of the sum of the exponential of each element of the input array.
+    //    And we use "rec_check" to ensure that the output is within the CQ range.
+    // 6. e^(X - m) * 1 / SUM(e^(X - m)):
+    //    We multiply the output from step 3 and step 5.
+    // 7. Change the scale factor of the output to the original scale factor.
     let max_output = graph.addNode(max, vec![(-1, 0)]);
     let sub_output = graph.addNode(sub, vec![(-1, 0), (max_output, 0)]);
     let exp_output = graph.addNode(exp, vec![(sub_output, 0)]);
--- a/src/layer/softmax.rs
+++ b/src/layer/softmax.rs
@@ -86,17 +86,18 @@ impl Layer for SoftmaxLayer {
     //    We first compute the maximum value of the input array. 
     // 2. X - m: 
     //    We subtract the maximum value from each element of the input array.
-    // 3. e^(X - m): 
+    // 3. e^(X - m) * SF: 
     //    We compute the exponential of each element of the input array. 
     //    And we use "exp_check" to ensure that the output is within the CQ range.
-    // 4. SUM(e^(X - m)): 
+    // 4. SUM(e^(X - m)) * SF: 
     //    We compute the sum of the exponential of each element of the input array.
-    // 5. 1 / SUM(e^(X - m)):
+    // 5. SF / SUM(e^(X - m)):
     //    We compute the reciprocal of the sum of the exponential of each element of the input array.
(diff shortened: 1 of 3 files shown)


src/layer/softmax.rs
```diff
@@ -42,14 +42,14 @@ impl Layer for SoftmaxLayer {
       basic_block: Box::new(SumBasicBlock {}),
       N: 1,
     }));
-    let log = graph.addBB(Box::new(LogBasicBlock {
+    let reciprocal = graph.addBB(Box::new(ReciprocalBasicBlock {
```
liliatangxy, line 45: how does this help with staying in CQ range?

src/util/onnx.rs
```diff
@@ -77,7 +77,7 @@ pub fn datumtype_to_sf(t: DatumType) -> usize {
     DatumType::I32 => 1,
     DatumType::I64 => 1,
     DatumType::Bool => 1,
-    DatumType::F32 => *onnx::SF,
+    DatumType::F32 => *onnx::SF_LOG,
```
liliatangxy, line 82: does this change affect layers that aren't nonlinear layers?

src/layer/softmax.rs
```diff
@@ -42,14 +42,14 @@ impl Layer for SoftmaxLayer {
...
       input_SF: *onnx::SF_LOG,
       output_SF: *onnx::SF_LOG,
     }));
-    let log_check = graph.addBB(Box::new(RepeaterBasicBlock {
+    let rec_check = graph.addBB(Box::new(RepeaterBasicBlock {
```
ddkang, line 49: Not clear what this does

src/util/onnx.rs
Vimalakirti: We only use this function for cast layer, so I think it won't. 
I added a comment to prevent any future misuse.

src/layer/softmax.rs
Vimalakirti: Thanks, I added comments in the layer to explain the proving idea.

src/layer/softmax.rs
Vimalakirti: Thanks, I added comments in the layer to explain the proving idea.
But let me explain a little more here:
Our goal is to compute Softmax(X) =  e^(X - m) * 1 / SUM(e^(X - m)), where m = max(X).
Given z = SUM(e^(X - m)), we used to compute a = Log(z), this step always succeeded in my experience. Then, we computed Softmax(X) = e^(X-m-a), this step was where a CQ range error often arose.
The reason is that some element in X-m is already a really really small negative number, X-m-a then easily falls out of the CQ range.
The new implementation uses Reciprocal to compute 1/z. (Remember z is not likely to fall out of CQ range.)
So we can compute Softmax(X) =  e^(X - m) * 1 / z successfully in the end.

liliatangxy (dismissed):

src/layer/softmax.rs
```diff
@@ -59,23 +59,55 @@ impl Layer for SoftmaxLayer {
...
     }));
 
+    // The proving idea is as follows
+    // 1. m = max(X): 
+    //    We first compute the maximum value of the input array. 
+    // 2. X - m: 
+    //    We subtract the maximum value from each element of the input array.
+    // 3. e^(X - m): 
+    //    We compute the exponential of each element of the input array. 
+    //    And we use "exp_check" to ensure that the output is within the CQ range.
```
ddkang, line 91: Why do the checks need to be in the proof? Won't they simply fail to prove?

src/layer/softmax.rs
```diff
@@ -59,23 +59,55 @@ impl Layer for SoftmaxLayer {
...
+    // 4. SUM(e^(X - m)): 
+    //    We compute the sum of the exponential of each element of the input array.
+    // 5. 1 / SUM(e^(X - m)):
```
ddkang: I'm pretty sure this and the next few steps are not efficient. Don't you want SF / sum? This is much more numerically stable

src/layer/softmax.rs
Vimalakirti: To prove computation for any nonlinear function y = N(x), we currently use a basicblock N to create y from x. 
However, the basicblock N has no proving, that's why we need to feed the pair (x, y) into the CQ2BB to prove it.
But I know it seems confusing, I raised an issue (https://github.com/uiuc-kang-lab/zk-torch/issues/125) to improve the readability by integrating CQ2 with N so that it will be more straightforward.

src/layer/softmax.rs
Vimalakirti: Sorry for my misleading comments. I've fixed the comments.
(We actually computed SF / sum)

ddkang (approved):

liliatangxy (approved):

PR #124, opened 2024-09-11
Add witness gen and load real input

Vimalakirti:
This PR added 
- witness generation under `bin/`
- loading real input from json. The input format is the same as that in `ezkl`


src/util/onnx.rs
```diff
@@ -72,6 +73,69 @@ pub fn datatype_to_datumtype(t: i32) -> DatumType {
...
+  let onnx = tract_onnx::onnx();
+  let onnx_graph = onnx.proto_model_for_path(onnx_name).unwrap().graph.unwrap();
+  let mut inputs = vec![];
+
+  let json = std::fs::read_to_string(json_name).expect("Failed to read file");
+  let json: InputData = serde_json::from_str(&json).unwrap();
+
+  for (i, onnx_input) in onnx_graph.input.iter().enumerate() {
+    let tract_onnx::pb::type_proto::Value::TensorType(t) = onnx_input.r#type.as_ref().unwrap().value.as_ref().unwrap();
+    let shape = t
```
liliatangxy: i think we might want to make this a separate function since it is used in many places

Vimalakirti: You mean the function to get the shape, right?

liliatangxy: yeah, i think it is helpful when we have to fix the dimensions when they are variable in the onnx file

Vimalakirti: I added a function so we can fix the variable dimension in util when we need to.

liliatangxy (dismissed):

ddkang (dismissed):

liliatangxy (approved):

PR #126, opened 2024-09-13
A binary search algorithm for finding opt sf

Vimalakirti:
This PR added the following features:
- implemented a binary search algorithm for finding the optimal scale factor.
- to achieve the above goal, also added `run` for CQ and CQ2 to check their ranges.
- (the editions in `softmax.rs` and `onnx.rs` are just re-formatting)

### DIFF
--- a/src/bin/search_sf.rs
+++ b/src/bin/search_sf.rs
@@ -1,47 +1,40 @@
 use core::panic;
-use std::fs;
-use std::process::{Command, Stdio};
-use zk_torch::{CONFIG, CONFIG_FILE};
+use zk_torch::{onnx, util, CONFIG};
 
-// Run the witness_gen binary and return the std output
-fn witness_gen() -> String {
-  let output = Command::new("./target/release/witness_gen")
-    .arg(CONFIG_FILE.clone())
-    .stdout(Stdio::piped()) // Capture stdout
-    .stderr(Stdio::piped()) // Capture stderr
-    .output() // Run the command and capture the output
-    .expect("Failed to execute command");
+// Run the witness_gen binary with a new sf and return the std output
+fn witness_gen(onnx_file_name: &str) -> String {
+  let (graph, models) = onnx::load_file(onnx_file_name);
 
-  // Capture stdout
-  let stdout = String::from_utf8_lossy(&output.stdout);
-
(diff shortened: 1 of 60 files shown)


liliatangxy (dismissed):

src/bin/search_sf.rs
```diff
@@ -0,0 +1,87 @@
+use core::panic;
+use std::fs;
+use std::process::{Command, Stdio};
+use zk_torch::{CONFIG, CONFIG_FILE};
+
+// Run the witness_gen binary and return the std output
+fn witness_gen() -> String {
+  let output = Command::new("./target/release/witness_gen")
```
ddkang: Don't use these random strings

src/bin/search_sf.rs
```diff
@@ -0,0 +1,87 @@
...
+  // Capture stderr (where panic messages typically appear)
+  let stderr = String::from_utf8_lossy(&output.stderr);
+  if !stderr.is_empty() {
+    println!("stdout msg:\n{}", stdout);
+    println!("stderr msg:\n{}", stderr);
+  }
+  format!("{}", stdout)
+}
+
+fn modify_config_sf(current_sf: usize, file_path: &str) {
```
ddkang: This seems like a really bad way to do this

ddkang: Is this a global SF?

src/bin/search_sf.rs
Vimalakirti: Removed

Vimalakirti:
The implementation allows us to search the optimal SF by running witness generation multiple times with different scale factors using the binary `search_sf`. Besides, the interfaces of functions in `search_sf.rs`  are now easier to understand. I recommend to run it with real/fixed input instead of random input (warning will pop out) because the optimal sf might fail when using another random input.

Specifically, the new commit catches the "CQ out of range error" by making `basicblock.run` and `graph.run` return `Result`. It also used a `RwLock` to wrap the `SF_LOG`, `SF`, and `SF_FLOAT` so that we can reset their values after initialization.

Please ignore all the edited files under `basic_block/` and `layer` as their functionalities are not changed.

liliatangxy (approved):

ddkang (approved):

PR #129, opened 2024-09-25
Fix InstanceNorm CQ range in vae decoder

liliatangxy:
- change to i128
- add the sqrt proving from SqrtLayer
- reduce the scaling needed in intermediate computations


Vimalakirti:
Could you please remove the unused `sqrt_check` based on CQ2 from `sqrt.rs`? 
Also, it will be nice if you can replace the sqrt check logics in BatchNorm.

Vimalakirti:
Thanks, please also remove the `sqrt_check` based on CQ2 from BatchnNorm.

src/util/arithmetic.rs
```diff
@@ -6,11 +6,12 @@
 use ark_bn254::Fr;
 use ark_ff::PrimeField;
 
-pub fn fr_to_int(x: Fr) -> i32 {
-  if x < Fr::from(1 << 28) {
-    x.into_bigint().0[0] as i32
+pub fn fr_to_int(x: Fr) -> i128 {
```
Vimalakirti, line 9: Is the "negative output issue" after Square resolved now?

src/basic_block/ops.rs
```diff
@@ -34,25 +34,25 @@ macro_rules! make_basic_block {
...
-make_basic_block!(TanhBasicBlock, { |x: f32| { x.tanh() } });
-make_basic_block!(CeilBasicBlock, { |x: f32| { x.ceil() } });
-make_basic_block!(NegBasicBlock, { |x: f32| { -x } });
-make_basic_block!(CosBasicBlock, { |x: f32| { x.cos() } });
-make_basic_block!(SinBasicBlock, { |x: f32| { x.sin() } });
-make_basic_block!(TanBasicBlock, { |x: f32| { x.tan() } });
-make_basic_block!(ReciprocalBasicBlock, { |x: f32| { 1. / x } });
+make_basic_block!(SqrtBasicBlock, { |x: f64| { x.sqrt() } });
+make_basic_block!(ChangeSFBasicBlock, { |x: f64| { x } });
+make_basic_block!(ErfBasicBlock, { |x: f64| { util::erf(x as f32) as f64 } });
```
Vimalakirti: Is there any reason why we don't just change the interface of `util::erf`?

src/util/arithmetic.rs
Vimalakirti: Sorry, I just checked the message in the channel. It is resolved.

liliatangxy: Yes, there are no more CQ out of range errors now

src/layer/norm.rs
```diff
@@ -186,8 +165,72 @@ impl Layer for BatchNormLayer {
...
+      output_SF: sf_log,
+    }));
+    let sf = onnx::SF.read().unwrap().to_owned();
+    let sf_const = graph.addBB(Box::new(Const2BasicBlock {
+      c: arr1(&vec![Fr::from(sf as i32)]).into_dyn(),
+    }));
+    let two_const = graph.addBB(Box::new(Const2BasicBlock {
+      c: arr1(&vec![Fr::from(2)]).into_dyn(),
+    }));
+    let add = graph.addBB(Box::new(RepeaterBasicBlock {
```
Vimalakirti: I think it will be better if we can remove redundant basicblocks (they are already defined before sqrt)

Vimalakirti (approved):

ddkang (approved):

PR #131, opened 2024-10-01
Karatsuba mul for high degree polynomials

liliatangxy:
Implements Karatsuba multiplication when the degree of the product exceeds the allowed degree from the evaluation domain (1<<28)

Timings with product of two polynomials of degree 2^x:
no karatsuba
x
20 duration: 695.99797ms
21 duration: 1.291637419s
22 duration: 2.808398087s
23 duration: 5.306105351s
24 duration: 9.263419673s
25 duration: 18.645103696s
26 duration: 37.43332432s
karatsuba
27 duration: 136.338719556s
28 duration: 356.805152415s
29 duration: 936.75454377s

PR #132, opened 2024-10-06
t poly split

liliatangxy:
Splits t poly commitment into deg < srs size polys to handle high degree t polys


Vimalakirti:
How's its security?

liliatangxy:
The main differences I'm aware of between this and that in the Plonk paper is the degree limit and that there isn't a fixed number of commitments of the t polynomial. I think these are ok because we are committing to the t commitments before getting the opening.

ddkang:
> The main differences I'm aware of between this and that in the Plonk paper is the degree limit and that there isn't a fixed number of commitments of the t polynomial. I think these are ok because we are committing to the t commitments before getting the opening.

Think is not good enough for security. Please produce a proof sketch

liliatangxy:
The proof for statistical zero knowledge is the same as that in the thesis Ari linked. In their case, they use the srs degree limit as the degree of the split, and they used a fixed number of t poly commitments. In our case, the number of t poly commitments is deterministic depending on the input size and srs size.

Here is the link to the thesis: https://repositum.tuwien.at/bitstream/20.500.12708/188597/1/Sefranek%20Marek%20-%202023%20-%20How%20to%20Simulate%20PLONK%20A%20Formal%20Security%20Analysis%20of%20a...pdf

src/util/poly.rs
```diff
@@ -52,3 +52,7 @@ pub fn mul_polys(polys: &Vec<DensePolynomial<Fr>>) -> DensePolynomial<Fr> {
   // Recursively call mul_polys on the next level until we get the root
   mul_polys(&next_level)
```
Vimalakirti, line 53: Do we still need #131?

liliatangxy: Yes since we still will need to do the full multiplication

Vimalakirti: Sure, then what's the proper order to merge them?

liliatangxy: They can be merged independently

Vimalakirti (approved):

ddkang:
Is there any performance degradation for no split?

liliatangxy:
No, tested on a resnet18 convolution

ddkang (approved):

PR #134, opened 2024-10-11
Add specialized concat for last dim

Vimalakirti:
This PR introduced a new concat basicblock. 
It can save 40%~50% of the proving time for concat operations and it is secure if our copy constraint is secure. 

The symbols used follow the ones in the copy constraint basic block, which is extended from Plonk paper.

### DIFF
--- a/src/basic_block/concat.rs
+++ b/src/basic_block/concat.rs
@@ -1,9 +1,492 @@
-use super::{BasicBlock, Data, DataEnc, PairingCheck, ProveVerifyCache, SRS};
-use crate::util;
-use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
-use ark_std::Zero;
-use ndarray::{Array1, ArrayD, Axis};
-use rand::rngs::StdRng;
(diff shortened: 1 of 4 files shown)


src/basic_block/concat.rs
```diff
@@ -1,9 +1,492 @@
-use super::{BasicBlock, Data, DataEnc, PairingCheck, ProveVerifyCache, SRS};
-use crate::util;
-use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
-use ark_std::Zero;
-use ndarray::{Array1, ArrayD, Axis};
-use rand::rngs::StdRng;
+#![allow(non_snake_case)]
```
ddkang, line 1: This seems very complicated. Does it share logic with other blocks? Can we make some of it modular?

Vimalakirti: Yeah, I agree copy constraint looks pretty complicated. I will look into ways that we can simplify the logics. For instance, the blinding replacement appears in other basicblocks, which could be modularized.

PR #137, opened 2024-10-22
Add mock setup

Vimalakirti:
This PR added the feature of mock setup, which allows us to benchmark proof generation without the needs to wait for real setup.


src/bin/mock_prove.rs
```diff
@@ -0,0 +1,142 @@
...
+  inputs: &Vec<&ArrayD<Fr>>,
+  graph: &Graph,
+  models: &Vec<&ArrayD<Fr>>,
+  timing: &mut TimingTree,
+) -> Result<Vec<Vec<ArrayD<Fr>>>, util::CQOutOfRangeError> {
+  // Run:
+  timed!(timing, "run witness generation", graph.run(inputs, models))
+}
+
+fn prove(
```
liliatangxy: is there a way to reuse the run/prove/main code from the other main?

Vimalakirti: yeah, we may put them under `util/`.
Please note that the `prove` in this file does not load setup from file as the `prove` in `main.rs` does, which means we will need to modify the `prove` in both files to make an unified API for them. Do you think this proposal make sense to you?

liliatangxy: right, i think we could maybe have a flag for mock vs. real and condition deserializing from file based on that

liliatangxy (dismissed):

src/basic_block.rs
```diff
@@ -169,6 +169,11 @@ pub trait BasicBlock: std::fmt::Debug + Send + Sync {
     (Vec::new(), Vec::new(), Vec::new())
   }
 
+  fn mockSetup(&self, _srs: &SRS, _model: &ArrayD<Data>) -> (Vec<G1Projective>, Vec<G2Projective>, Vec<DensePolynomial<Fr>>) {
```
ddkang: Is there a way to just infer this directly from the type signature? This seems pretty inelegant

Vimalakirti: yeah, the new commit addressed this issue.

ddkang (approved):

liliatangxy (approved):

PR #140, opened 2024-10-26
Transpose+permute optimization for 1x1 kernel

liliatangxy:
Instead of using copy constraint for a 1x1 kernel, we can use transpose+permute for ch_in to be the last dimension for the matmul.

first 1x1 kernel layer in resnet50 old -> new: 932s -> 651s (-30%)

### DIFF
--- a/src/layer/conv.rs
+++ b/src/layer/conv.rs
@@ -98,7 +98,7 @@ fn splat_input(
   inp_cells
 }
 
-// Splats weights into shape (out_channels x inp_channels * kernel_dims product) such that each row corresponds to flattened kernels for each input channel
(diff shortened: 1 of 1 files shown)


Vimalakirti:
Do you have estimates on how much it could improve?

liliatangxy:
updated the timings

Vimalakirti (commented):
Seems good for me, but it looks pretty complicated. Do you think it will look better if we group all 1x1 computations together under one if block?

liliatangxy:
I think grouping in one if might have a lot of repeated code. I added additional comments and grouped the optimization code together more, hope that helps.

Vimalakirti (approved):

ddkang (approved):

PR #141, opened 2024-10-29
Add multihead matmul

Vimalakirti:


src/layer/matmul.rs
```diff
@@ -42,9 +41,7 @@ impl Layer for MatMulLayer {
       }),
       N: 1,
     }));
-    let small_matmul = (a * b) < 1 << CONFIG.ptau.loaded_pow_len_log;
-    let use_cqlin = constants.len() > 1 && constants[1].is_some() && small_matmul;
-    let matmul_output = if use_cqlin {
+    let matmul_output = if constants.len() > 1 && constants[1].is_some() {
```
ddkang: What's going on here?

Vimalakirti: Good catch. I rebased on an incorrect commit. 
Now it's fixed.

ddkang:
Can we make a test for this?

Vimalakirti:
> Can we make a test for this?

We only have tests for basicblocks. If we want to test layers, we need to implement other logics to support it. But yes, we can test it later.

liliatangxy (approved):

ddkang (approved):

PR #142, opened 2024-10-29
Fix Conv output permute

liliatangxy:


ddkang:
What was the issue?

Do we have tests for this?

liliatangxy:
The permute for the output used the input channel dimension instead of the output channel dimension

I tested it on the resnet50 onnx file. Before I tested it only on a single layer and didn't check the output shape carefully

ddkang:
Can we make a simple test for this?

PR #143, opened 2024-10-30
Change reshape ccbb to permute

liliatangxy:
Changes reshape copy constraint bb into reshape+transpose+permute bbs in Conv and Maxpool. A similar change can possibly be made to other layers which use reshape operations.


ddkang:
Does this speed things up? If so, how much?

liliatangxy:
This replaces one copy constraint for each of these layers which proves on the order of minutes to something on the order of seconds. I queued resnet50 now

PR #148, opened 2024-11-10
Fixes for vae-decoder and unet

liliatangxy:
- Enable ReshapeBB to work when converting to view doesn't work
- Fix pow of 2 issues in old Conv, Mul, InstanceNorm
- Optimize out copy constraint when reshape is possible in old Conv, InstanceNorm, and MaxPool which are used in vae and unet
- Fix end indexing in Slice


ddkang:
What do these fixes do?

PR #149, opened 2024-11-17
New conv and maxpool

Vimalakirti:
This PR introduced a new Conv2D and MaxPool2D without any copy constraint.
The key idea here is viewing convolution as matrix multiplication so that we can use only cqlins to prove it.
And when kernel size is 1x1, the convolution can be implemented with only one cqlin and two permutes.

MaxPool2D is similar, the window sliding process can be viewed as matrix multiplication.  I used cq checks to ensure the MaxPool result is not less than any element in the corresponding window. Besides, I added selectors to ensure the MaxPool elements exist in the original input.

(The editions in `graph.rs` are used to truncate the basicblock info. in our log, it will be moved to other PR later.)

There are more things to be done:
- [x] Given kernel size = [c_o, c_i, h, w], we do `c_i * c_o` times matrix multiplications in one Conv2D. But I suspect `c_o` times matrix multiplication + 2 permutes could be faster (if I merge the c_i channels into matmul). I will benchmark this and see if we need to implement in this way.
- [x] Provide benchmark results on resnet-50 (I should provide the stats by today, but the results will not be ready until midnight.  Let me show the results here tomorrow.)
- [x] Refactor the codes and add comments for readability
- [x] I'm also thinking that this efficient Conv may help us prove 3DUnet. To achieve this, I will implement new Conv3D and ConvTranspose3D. (will be in another PR if we find a more efficient way)
- [x] Add the python code to edit the resnet onnx


Vimalakirti:
Benchmarks:
ResNet50 (Main branch)
- Proving time: 25417.27s
- Verification time: 1656.07s
- Proof size: 1.11GB

ResNet50 (This branch)
- Proving time: 11216.97s
- Verification time: 2086.59s
- Proof size: 1.42GB

Currently, if kernel is not [1,1], I do C_in * C_out CQLins to transform each input slice [H * W] to the corresponding output slice [H' * W'].
I think if we view each Conv2D as **one matrix multiplication** that maps input [1, C_in * H * W] to [1, C_out * H' * W'], we only need one CQLin per layer, which should decrease the proof size a lot. I will design and implement this.

ddkang:
>  I'm also thinking that this efficient Conv may help us prove 3DUnet. To achieve this, I will implement new Conv3D and ConvTranspose3D.

Before you do this, we need to check that the 1st conv doesn't OOM otherwise it will be a lot of effort for something we can't prove anyway

src/basic_block/cqlin.rs
```diff
@@ -9,6 +9,384 @@ use ark_std::{UniformRand, Zero};
 use ndarray::{ArrayD, Ix2, IxDyn};
 use rand::{rngs::StdRng, SeedableRng};
 use rayon::prelude::*;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+fn get_model_element(kernel: &ArrayD<i128>, indices: &HashMap<(usize, usize), KernelIdx>, i: usize, j: usize) -> Fr {
```
ddkang, line 15: What does this do?

src/layer/new_conv.rs
```diff
@@ -0,0 +1,353 @@
...
+use ndarray::{s, Array1, Array4, ArrayD, Axis};
+use rayon::prelude::*;
+use std::collections::HashMap;
+use std::sync::Arc;
+use tract_onnx::pb::AttributeProto;
+use tract_onnx::prelude::DatumType;
+use tract_onnx::prelude::Outlet; // Import Rayon traits
+
+// Return the "indices" field in SparseCQLinBasicBlock for a given input and kernel shape
+pub fn get_kernel_map(
```
ddkang, line 24: What does "kernel_map" mean?

src/layer/new_conv.rs
```diff
@@ -0,0 +1,353 @@
...
+        }
+      }
+    }
+  });
+
+  (kernel_hashmap, total_outputs)
+}
+
+// Slice a kernel map to a smaller kernel map
+pub fn slice_kernel_map(
```
ddkang, line 87: What is this used for?

src/layer/new_conv.rs
```diff
@@ -0,0 +1,353 @@
+use crate::basic_block::*;
```
ddkang, line 1: These files are incredibly complicated. Why are they so complex?

PR #151, opened 2024-12-10
Batch CQ

liliatangxy:
Batch CQ draft. Requires a total of three passes to avoid saving all input polynomials, one to commit all the inputs, one for non-opening commitments, one for opening commitments, but currently, these three passes are done in addition to the normal prove pass.

- [ ] Add serialization for commitments between rounds for Fiat-Shamir rng
- [ ] Fix rng correction factors (Cs) at proof
- [x] Add graph batch verify
- [x] Add batch prove/verify to main

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -17,6 +17,7 @@ prover:
   enc_input_path: inputsEnc
   enc_output_path: outputsEnc
   proof_path: proofs
+  batch_proof_path: batch_proofs
   enable_layer_setup: true
 verifier:
   enc_model_path: modelsEnc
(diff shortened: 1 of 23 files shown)


Vimalakirti:
What should we change in layers to integrate the new CQ?

liliatangxy:
As it is now, we don't need to change layers

Vimalakirti:
Thanks for your efforts on this PR! I read `graph.rs` and `cq.rs`. I think it'll be more useful if batch cq can batch all cq/cq2 for polynomials with the same degree and the same type of basicblock across the entire graph. For example, batch degree-n input and output polys for all ReLU across the model. But the batch cq seems to only batch the cq for the same input and output indices. Am I understanding wrong?

liliatangxy:
The key for `BatchProveState` across different input/output indices with the same degree should be the same if the CQ basic block `CQArrayType` is the same.

liliatangxy:
updated in #152

PR #153, opened 2025-03-09
Mira CQLin

Vimalakirti:
This branch includes the progress to integrate Mira into zk-torch. It includes the following features:
- New fns in BB that allows us to do piecemeal development.
- New graph.prove and graph.verify to integrate the Mira folding scheme. The old prove() and verify() without folding is still available when "--features=fold" is off.
- A unifying Accumulator data holder defined in util/fold.rs (This reduces the difficulty to read the code)
- Update CQLin without high-degree gamma^n by the Mira optimization
- Integrating Mira CQLin into zk-torch (working on it):
  - My previous implementation for Mira CQLin (includes runnable and correct prover and decider)(https://drive.google.com/drive/folders/1WT2TUw2wV0lMbk_6bWHbsdzOsreV97_n?usp=sharing)

If the first three items look good, we can work on Mira folding for different BBs in parallel.

My TODOs for this branch: 
- [x] RLC checks in CQLin acc_verify
- [x] Fiat-shamir in CQLin Acc
- [x] Fix the index in the data holder of CQLin accumulator.
- [x] Fold all the CQLins with the same matrix shapes


src/basic_block.rs
```diff
@@ -196,4 +196,57 @@ pub trait BasicBlock: std::fmt::Debug + Send + Sync {
...
+    _outputs: &Vec<&ArrayD<Data>>,
+    _acc_proof: (&Vec<G1Projective>, &Vec<G2Projective>, &Vec<Fr>),
+    _proof: (&Vec<G1Projective>, &Vec<G2Projective>, &Vec<Fr>),
+    _rng: &mut StdRng,
+    _cache: ProveVerifyCache,
+  ) -> (Vec<G1Projective>, Vec<G2Projective>, Vec<Fr>) {
+    (Vec::new(), Vec::new(), Vec::new())
+  }
+
+  // This function cleans the blinding terms in accumulators for the verifier to do acc_verify
```
ddkang, line 230: Is clean the right word? What do people say in the literature?

src/util/fold.rs
```diff
@@ -0,0 +1,88 @@
+use ark_bn254::Fr;
+
+pub struct AccHolder<P: Clone, Q: Clone> {
```
ddkang, line 15: Does this work in general?

src/basic_block.rs
Vimalakirti: There's no acc.clean in the literature since they don't design and implement zero knowledge (neither ProtoStar nor Mira). We need this function because we need CQLin.prove to return additional blinding factors to do the folding correctly; however, we don't want verifier to see these terms, so I add this fn for this purpose.

src/util/fold.rs
Vimalakirti: Yeah, it's a general accumulator data holder for all basicblocks.

src/basic_block.rs
Vimalakirti: (To make it aligned with the literature, we can wrap acc.clean into acc.prove in the paper)

Vimalakirti:
All the TODOs are done, and the PR is ready for review.
Besides the above, I added some new features including:
- Update util::combine_pairing_checks() to support infinity points
- A test module in tests.rs to test folding. (Thanks for lilia's suggestion)
- The fn graph.prove() will automatically fold all the foldable acc_proofs (e.g., CQLins whose matrices are in the same shape).
  - This relies a fn called `get_foldable_bb_info` in `util/fold.rs` to summarize BB information for folding such that we can aggregate proofs for all foldable basicblocks. It should be updated incrementally later when new BB foldings are implemented.

I will continue working on new branches for MatMul, Permute, and Mul from this branch.

src/basic_block.rs
```diff
@@ -14,6 +14,7 @@ pub use cq::CQBasicBlock;
 pub use cq2::CQ2BasicBlock;
 pub use cqlin::CQLinBasicBlock;
 pub use div::{DivConstBasicBlock, DivConstProofBasicBlock, DivScalarBasicBlock, ModConstBasicBlock};
+use downcast_rs::impl_downcast;
```
ddkang, line 17: What is this for?

src/onnx.rs
```diff
@@ -419,5 +421,17 @@ pub fn load_file(filename: &str) -> (Graph, Vec<(ArrayD<Fr>, DatumType)>) {
 
   propagate_precomputable(&mut graph);
 
+  #[cfg(feature = "fold")]
+  {
+    let mut bb_to_index = HashMap::new();
+    let mut foldable_bb_map = HashMap::new();
+    for (i, b) in graph.basic_blocks.iter().enumerate() {
+      let bb_info_for_folding = get_foldable_bb_info(&b);
```
ddkang: Is this the best way to do this?

src/util/fold.rs
```diff
@@ -0,0 +1,99 @@
...
+  acc.acc_errs.iter().for_each(|x| {
+    g1.extend(x.0.clone());
+    g2.extend(x.1.clone());
+    fr.extend(x.2.clone());
+  });
+  fr.push(acc.mu);
+  (g1, g2, fr)
+}
+
+pub fn acc_proof_to_cqlin_acc<P: Clone, Q: Clone>(acc_proof: (&Vec<P>, &Vec<Q>, &Vec<Fr>), log_n: usize, is_prover: bool) -> AccHolder<P, Q> {
```
ddkang: It's kind of weird that there's a cqlin specific function here. This seems like bad organization / architecture

src/util/prover.rs
```diff
@@ -289,7 +289,7 @@ pub fn zktorch_kernel() {
   #[cfg(not(feature = "mock_prove"))]
   verify(&srs, &graph, &mut timing);
 
-  // Measure proof size
+  // Measure proof size; TODO: in folding mode, the proof size is not accurate
```
ddkang: How do you fix this? This is important

ddkang (commented):
Make sure to push the other basic blocks in a different branch

src/basic_block.rs
Vimalakirti: This is a convenient way to downcast `dyn BasicBlock` to a specific BB such as CQLin, which allows us to indicate those foldable BBs.

src/util/fold.rs
Vimalakirti: Good point, I moved the fn from here to cqlin.rs

src/util/prover.rs
Vimalakirti: My latest commit addresses this issue. 
The new proof size includes:
1) the proofs from those non-foldable BBs (some of them will be made foldable in few days)
2) the acc proofs from those foldable BBs without counting the size of `errs` as they are not needed for the final acc decider (that's why we need a new fn `acc_clean_errs` in BBs)

src/onnx.rs
Vimalakirti: I refactored it a little, but yeah, this is probably the most convenient way I can think of to identify foldable BBs without hurting the prover and verifier time.

ddkang (commented):
Please push the next simplest basic block so we can stress test the API and we can merge after

Vimalakirti:
> Please push the next simplest basic block so we can stress test the API and we can merge after

Sure, the latest commit includes folding for MulConst.

ddkang (approved):

liliatangxy (approved):

PR #154, opened 2025-03-16
Fold Mulscalar, Mul, and Sum

Vimalakirti:
This PR includes folding for the `mulscalar`, `mul`, and `sum` protocols.

In addition to folding, I made some updates to the layers and tests because:

1. The length of the vector is required for `mul` and `sum` to perform folding.
2. `util::combine_pairing_check` occasionally fails for unknown reasons. As a workaround, I replaced it with the primitive multi_pairing from Arkworks. We can revisit and address the issue with `combine_pairing_check` in a future PR.


ddkang:
Can we run some benchmarks at this point?

Vimalakirti:
> Can we run some benchmarks at this point?

We can, but I suggest we run benchmarks when Mira CQ is done since it was the main bottleneck of the proof size in CNNs

src/basic_block/mul.rs
```diff
@@ -201,10 +290,273 @@ impl BasicBlock for MulScalarBasicBlock {
...
+    let [acc_inp1_2] = acc_holder.acc_g2[..] else {
+      panic!("Wrong acc proof format")
+    };
+    let acc_mu = acc_holder.mu;
+    let [acc_inp0_r, acc_inp1_r] = acc_holder.acc_fr[..] else {
+      panic!("Wrong acc proof format")
+    };
+
+    // Compute the error
+    let err: (Vec<G1Projective>, Vec<G2Projective>, Vec<Fr>) = (
```
liliatangxy, line 370: Can we leave comments with the modified checks and error terms for all BBs?

liliatangxy (dismissed):

src/basic_block/mul.rs
```diff
@@ -201,10 +290,273 @@ impl BasicBlock for MulScalarBasicBlock {
...
+      vec![acc_inp1_2, inp1_2],
+      vec![],
+    );
+    let mut errs = vec![err];
+
+    // Fiat-Shamir
+    let mut bytes = Vec::new();
+    acc_holder.acc_g1[1..4].serialize_uncompressed(&mut bytes).unwrap();
+    acc_holder.acc_g2.serialize_uncompressed(&mut bytes).unwrap();
+    proof.0[1..proof.0.len() - 4].serialize_uncompressed(&mut bytes).unwrap();
```
liliatangxy: I think this line and the corresponding line in acc_verify serialize an empty slice

Vimalakirti: removed

src/basic_block/mul.rs
Vimalakirti: Left new comments for errs

Vimalakirti:
@ddkang @liliatangxy please see if this PR can be merged

ddkang (approved):

PR #155, opened 2025-03-18
Fold MatMul and Permute

Vimalakirti:
This PR supports folding for MatMul and Permute.

PR #159, opened 2025-03-27
Mira readability

liliatangxy:
- Improve readability and simplify fiat-shamir and prover/verifier field logic with named structs for accumulator 
- Contains matmul only for now, working on the same for cq, cq2

### DIFF
--- a/src/basic_block/matmul.rs
+++ b/src/basic_block/matmul.rs
@@ -11,6 +11,291 @@ use ndarray::{arr1, arr2, ArrayD, Ix1, Ix2, IxDyn};
 use rand::{rngs::StdRng, SeedableRng};
 use rayon::iter::ParallelIterator;
 
+struct MatMulAccProof<P: Clone + CanonicalSerialize, Q: Clone + CanonicalSerialize> {
+  fiat_shamir: MatMulAccFiatShamir<P, Q>,
+  acc_corr1: P,
+  acc_corr2: P,
+  acc_corr3: P,
+  acc_corr4: P,
+  mu: Fr,
+  prover_only: Option<MatMulAccProofProverOnly<P>>,
+  errs: MatMulErrs<P, Q>,
+  acc_errs: MatMulErrs<P, Q>,
+}
+
+#[derive(CanonicalSerialize)]
+struct MatMulAccFiatShamir<P: Clone + CanonicalSerialize, Q: Clone + CanonicalSerialize> {
+  acc_left_x: P,
+  acc_left_Q_x: P,
+  acc_left_zero: P,
+  acc_left_zero_div: P,
+  acc_right_x: P,
+  acc_right_Q_x: P,
+  acc_right_zero_div: P,
+  acc_flat_A: P,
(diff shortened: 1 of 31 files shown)


Vimalakirti:
It might be better to merge readability stuffs into #155  as we need to merge that branch first.

src/tests.rs
```diff
@@ -196,15 +196,15 @@ fn testBasicBlocks() {
...
-  testBasicBlock(
-    RepeaterBasicBlock {
-      basic_block: Box::new(MulBasicBlock { len: N }),
-      N: 1,
-    },
-    srs,
-    &empty,
-    &vec![&A, &B],
-  );
+  // testBasicBlock(
```
Vimalakirti, line 199: What is this for?

liliatangxy: The test was failing and has nothing to do with this PR.

Vimalakirti:
Thanks! Is it possible to include only the readability feature by removing the features of `mira_cq` and `mira_repeat`?

src/basic_block/matmul.rs
```diff
@@ -11,45 +11,323 @@ use ndarray::{arr1, arr2, ArrayD, Ix1, Ix2, IxDyn};
...
+    flat_A: errs.flat_A * acc_gamma,
+    acc_flat_A: errs.acc_flat_A * acc_gamma,
+    acc_left_Q_x: errs.acc_left_Q_x * acc_gamma,
+    acc_left_x: errs.acc_left_x * acc_gamma,
+    acc_part_corr1: errs.acc_part_corr1 * acc_gamma,
+    acc_flat_B_g2: errs.acc_flat_B_g2,
+    flat_B_g2: errs.flat_B_g2,
+  };
+
+  let acc_errs = MatMulErrs {
```
Vimalakirti, line 90: I think this won't pass the checks in decider since old flat_A, flat_B_g2, acc_flat_A, and acc_flat_B_g2 are not included in the final error.

src/basic_block/matmul.rs
```diff
@@ -593,65 +828,68 @@ impl BasicBlock for MatMulBasicBlock {
...
+    let errs = &acc.acc_errs;
 
     let mut temp: PairingCheck = vec![];
-    for i in 0..err_1.1.len() {
-      temp.push((-err_1.0[i], err_1.1[i]));
-    }
-    temp.push((err_1.0[err_1.1.len()], (srs.X2A[self.m] - srs.X2A[0]).into()));
-    temp.push((err_1.0[err_1.1.len() + 1], srs.X2A[0]));
-    temp.push((err_1.0[err_1.1.len() + 2], srs.Y2A));
+    temp.push((-errs.flat_A, errs.acc_flat_B_g2));
```
Vimalakirti, line 888: This won't pass, please see the reason above.

src/tests.rs
Vimalakirti: Thanks, it might be a bug in `mira_repeat`. If we remove `mira_repeat` and `mira_cq` features, we will not need these.

PR #160, opened 2025-03-27
Mira matmul/permute readability

liliatangxy:
Sample readability for matmul and permute to include named structs that also simplify logic for what is included in fiat-shamir and prover vs. verifier.

Bing-Jyue says that the parallelization features will change the API, so the design may change again based on that.

### DIFF
--- a/src/basic_block/matmul.rs
+++ b/src/basic_block/matmul.rs
@@ -13,10 +13,7 @@ use rayon::iter::ParallelIterator;
 
 struct MatMulAccProof<P: Clone + CanonicalSerialize, Q: Clone + CanonicalSerialize> {
   fiat_shamir: MatMulAccFiatShamir<P, Q>,
-  acc_corr1: P,
-  acc_corr2: P,
-  acc_corr3: P,
-  acc_corr4: P,
+  acc_corr: [P; 4],
   mu: Fr,
   prover_only: Option<MatMulAccProofProverOnly<P>>,
   errs: MatMulErrs<P, Q>,
@@ -114,10 +111,12 @@ fn accumulate(
       acc_flat_B_g2: matmul_acc.fiat_shamir.acc_flat_B_g2 + flat_B_g2 * acc_gamma,
       acc_beta_pow_g2: beta_pow_g2,
     },
-    acc_corr1: matmul_acc.acc_corr1 + corr1 * acc_gamma,
-    acc_corr2: matmul_acc.acc_corr2 + corr2 * acc_gamma,
-    acc_corr3: matmul_acc.acc_corr3 + corr3 * acc_gamma,
-    acc_corr4: matmul_acc.acc_corr4 + corr4 * acc_gamma,
+    acc_corr: [
+      matmul_acc.acc_corr[0] + corr1 * acc_gamma,
+      matmul_acc.acc_corr[1] + corr2 * acc_gamma,
+      matmul_acc.acc_corr[2] + corr3 * acc_gamma,
+      matmul_acc.acc_corr[3] + corr4 * acc_gamma,
+    ],
     mu: matmul_acc.mu + acc_gamma,
     prover_only: Some(MatMulAccProofProverOnly {
       acc_part_corr1: matmul_acc_prover_only.acc_part_corr1 + part_corr1 * acc_gamma,
@@ -205,10 +204,12 @@ fn matmul_acc_holder_to_acc<P: Clone + CanonicalSerialize, Q: Clone + CanonicalS
       acc_flat_B_g2: acc_holder.acc_g2[0].clone(),
       acc_beta_pow_g2: acc_holder.acc_g2[1].clone(),
     },
-    acc_corr1: acc_holder.acc_g1[7].clone(),
-    acc_corr2: acc_holder.acc_g1[8].clone(),
-    acc_corr3: acc_holder.acc_g1[9].clone(),
-    acc_corr4: acc_holder.acc_g1[10].clone(),
+    acc_corr: [
+      acc_holder.acc_g1[7].clone(),
+      acc_holder.acc_g1[8].clone(),
+      acc_holder.acc_g1[9].clone(),
+      acc_holder.acc_g1[10].clone(),
+    ],
     mu: acc_holder.mu,
     prover_only: if is_prover {
       Some(MatMulAccProofProverOnly {
@@ -281,10 +282,10 @@ fn matmul_acc_to_acc_holder<P: Clone + CanonicalSerialize, Q: Clone + CanonicalS
       acc.fiat_shamir.acc_right_x,
       acc.fiat_shamir.acc_right_Q_x,
       acc.fiat_shamir.acc_right_zero_div,
-      acc.acc_corr1,
-      acc.acc_corr2,
-      acc.acc_corr3,
-      acc.acc_corr4,
+      acc.acc_corr[0].clone(),
+      acc.acc_corr[1].clone(),
+      acc.acc_corr[2].clone(),
+      acc.acc_corr[3].clone(),
       acc.fiat_shamir.acc_flat_A,
       acc.fiat_shamir.acc_flat_B,
       acc.fiat_shamir.acc_flat_C,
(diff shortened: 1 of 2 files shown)


src/basic_block/matmul.rs
```diff
@@ -11,7 +11,129 @@ use ndarray::{arr1, arr2, ArrayD, Ix1, Ix2, IxDyn};
...
+    acc_flat_A: errs.acc_flat_A * acc_gamma,
+    acc_left_Q_x: errs.acc_left_Q_x * acc_gamma,
+    acc_left_x: errs.acc_left_x * acc_gamma,
+    acc_part_corr1: errs.acc_part_corr1 * acc_gamma,
+    acc_flat_B_g2: errs.acc_flat_B_g2,
+    flat_B_g2: errs.flat_B_g2,
+  };
+
+  let acc_errs = MatMulErrs {
+    flat_A: new_errs.flat_A,
```
Vimalakirti, line 102: (The same as the comments in #159) The old `matmul_acc.flat_A`, `matmul_acc.acc_flat_A`, `matmul_acc.flat_B`, `matmul_acc.acc_flat_B` should be included in the new acc_errs. Or they will not be verified by the decider.

src/basic_block/matmul.rs
```diff
@@ -11,7 +11,129 @@ use ndarray::{arr1, arr2, ArrayD, Ix1, Ix2, IxDyn};
 use rand::{rngs::StdRng, SeedableRng};
 use rayon::iter::ParallelIterator;
 
-fn acc_proof_to_matmul_acc<P: Clone, Q: Clone>(acc_proof: (&Vec<P>, &Vec<Q>, &Vec<Fr>), is_prover: bool) -> AccHolder<P, Q> {
+struct MatMulAccProof<P: Clone + CanonicalSerialize, Q: Clone + CanonicalSerialize> {
+  fiat_shamir: MatMulAccFiatShamir<P, Q>,
+  acc_corr1: P,
```
Vimalakirti: Would it be better to make `acc_corrs` a Vec?

src/basic_block/matmul.rs
liliatangxy: Sorry, I'm not sure if I understand. Is the decider missing these checks currently?

src/basic_block/matmul.rs
liliatangxy: made it an array to fix the number of elements

src/basic_block/matmul.rs
Vimalakirti: No worries, let's look into this closely:
acc_left_Q_x, acc_left_x, and acc_part_corr1 are combination of matmul_acc.XXX + new_errs.XXX
Then, why acc_flat_A and acc_flat_B are not?

liliatangxy: I thought that in the original code they are not added, please let me know if I am missing something.

```rust
new_acc_holder.acc_errs = acc_holder.acc_errs;

// this is what new_errs has
// so (new_errs.flat_A, new_errs.acc_flat_A) = (errs.flat_A * acc_gamma, errs.acc_flat_A * acc_gamma)
errs[0].0 = errs[0].0.iter().map(|x| (*x * acc_gamma).into()).collect();

let err1_g1_len = new_acc_holder.acc_errs[0].0.len();
let q_term_g1 = new_acc_holder.acc_errs[0].0[err1_g1_len - 3].clone() + errs[0].0[2];
let l_term_g1 = new_acc_holder.acc_errs[0].0[err1_g1_len - 2].clone() + errs[0].0[3];
let c_term_g1 = new_acc_holder.acc_errs[0].0[err1_g1_len - 1].clone() + errs[0].0[4];
// this is equal to vec![new_errs.flat_A, new_errs.acc_flat_A]
let mut errs_0_g1 = errs[0].0[..2].to_vec();
let mut errs_0_g2 = errs[0].1[..2].to_vec();

// this is acc_err
new_acc_holder.acc_errs[0].0 = new_acc_holder.acc_errs[0].0[..err1_g1_len - 3].to_vec();
// (acc_err.flat_A, acc_err.acc_flat_A) = (new_errs.flat_A, new_errs.acc_flat_A)
new_acc_holder.acc_errs[0].0.append(&mut errs_0_g1);
new_acc_holder.acc_errs[0].0.push(q_term_g1);
new_acc_holder.acc_errs[0].0.push(l_term_g1);
new_acc_holder.acc_errs[0].0.push(c_term_g1);
new_acc_holder.acc_errs[0].1.append(&mut errs_0_g2);
```

Likewise, in the original code only q, l, c are only checked in the verifier
```rust
acc_holder.errs[0].0[acc_holder.errs[0].0.len() - 3..]
      .iter()
      .zip(prev_acc_holder.acc_errs[0].0[prev_acc_holder.acc_errs[0].0.len() - 3..].iter())
      .enumerate()
      .for_each(|(j, (x, y))| {
        let z = *y + *x * acc_gamma;
        result &= z == acc_holder.acc_errs[0].0[acc_holder.acc_errs[0].0.len() - 3 + j];
      });
```

Vimalakirti: Thank you for referencing the code. Let me show you they were actually added (appended).
```Rust
// this line copies the original acc_errs
new_acc_holder.acc_errs = acc_holder.acc_errs;
...
// IMPORTANT: this line includes all terms related to acc_flat_A and acc_flat_B from the previous acc_errs
new_acc_holder.acc_errs[0].0 = new_acc_holder.acc_errs[0].0[..err1_g1_len - 3].to_vec();
// this line appends the new terms related to acc_flat_A and acc_flat_B in the latest acc_err
new_acc_holder.acc_errs[0].0.append(&mut errs_0_g1);
// the following lines includes Q, L, C terms
new_acc_holder.acc_errs[0].0.push(q_term_g1);
new_acc_holder.acc_errs[0].0.push(l_term_g1);
new_acc_holder.acc_errs[0].0.push(c_term_g1);
```

liliatangxy: Thanks, I accumulate all the old terms with a vec since it increases each round and I don't think it hurts readability much. Please let me know if it looks ok

src/basic_block/matmul.rs
```diff
@@ -53,6 +193,174 @@ fn acc_proof_to_matmul_acc<P: Clone, Q: Clone>(acc_proof: (&Vec<P>, &Vec<Q>, &Ve
...
+  is_prover: bool,
+) -> Option<MatMulAccProof<P, Q>> {
+  if acc_proof.0.len() == 0 && acc_proof.1.len() == 0 && acc_proof.2.len() == 0 {
+    return None;
+  }
+  let acc_holder = acc_proof_to_matmul_acc_holder(acc_proof, is_prover);
+  Some(matmul_acc_holder_to_acc(acc_holder, is_prover))
+}
+
+fn matmul_acc_holder_to_acc<P: Clone + CanonicalSerialize, Q: Clone + CanonicalSerialize>(
```
Vimalakirti: To represent Acc, we now have three different ways: 
1. AccProof: similar to proof, which is `(&Vec<G1Projective>, &Vec<G2Projective>, &Vec<Fr>)`
2. AccHolder: a general acc data holder for all acc, it is defined in `util`
3. MatMulAccProof: added in this PR

I feel this is complex because we have multiple fn XXX_to_OOO, where XXX and OOO could be any pair from the above three choices. Is there any way to make it simpler?

src/basic_block/matmul.rs
```diff
@@ -53,6 +193,174 @@ fn acc_proof_to_matmul_acc<P: Clone, Q: Clone>(acc_proof: (&Vec<P>, &Vec<Q>, &Ve
...
+  acc_holder: AccHolder<P, Q>,
+  is_prover: bool,
+) -> MatMulAccProof<P, Q> {
+  let g1_len = acc_holder.acc_errs[0].0.len();
+  let g2_len = acc_holder.acc_errs[0].1.len();
+  let acc_proof = MatMulAccProof {
+    fiat_shamir: MatMulAccFiatShamir {
+      acc_left_x: acc_holder.acc_g1[0].clone(),
```
Vimalakirti: Btw, this way also requires us to clone our data frequently, which might hurt the performance. Is there any way that we can prevent this?

src/basic_block/matmul.rs
liliatangxy: I agree it is many transitions. Ideally we could avoid AccProof or AccHolder entirely, but if we remove AccHolder, I think we leave a lot of implementation details to be repeated in the conversion to and from the BB-specific structs. 
If I understand correctly, I think conversion between AccHolder and AccProof can just be done in graph and doesn't need to be touched by the BB implementations. If this is true, it's better to introduce that in a separate PR.

src/basic_block/matmul.rs
liliatangxy: Changed everything to use Copy instead of Clone trait, which should avoid clone()

Vimalakirti (approved):

ddkang (commented):
Much better. Are there any performance regressions?

liliatangxy:
It doesn't seem so. Benched with 1x512, 512x512 matmul and 256x256 permute

This change occurs in acc proof:
old matmul
Proof 1 time: 94.799237ms
Proof 2 time: 72.919111ms
Acc proof time: 4.311593ms

new matmul
Proof 1 time: 92.245745ms
Proof 2 time: 70.669219ms
Acc proof time: 4.051291ms

old permute
Proof 1 time: 5.048765434s
Proof 2 time: 2.546416728s
Acc proof time: 1.977149ms

new permute
Proof 1 time: 5.066130832s
Proof 2 time: 2.529141491s
Acc proof time: 2.021624ms

ddkang:
lgtm

PR #161, opened 2025-04-01
Setup deduplication

liliatangxy:
A lot of memory is wasted on duplicate setup vec elements. This enables storing these elements exactly once between cq, cq2, cqlin.

Note that when running tests going forward, we want to use the `--release` flag since the `SetupCache` `Arc<Mutex<..>>` is slow otherwise.

### DIFF
--- a/src/basic_block/cqlin.rs
+++ b/src/basic_block/cqlin.rs
@@ -1,6 +1,6 @@
 #![allow(non_snake_case)]
 #![allow(non_upper_case_globals)]
-use super::{BasicBlock, CacheValues, Data, DataEnc, PairingCheck, ProveVerifyCache, SRS};
+use super::{BasicBlock, CacheValues, Data, DataEnc, PairingCheck, ProveVerifyCache, SetupCache, SRS};
 use crate::util::{self, acc_to_acc_proof, calc_pow, AccHolder};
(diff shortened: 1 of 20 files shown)


ddkang:
> A lot of memory is wasted on duplicate setup vec elements. This enables storing these elements exactly once between cq, cq2, cqlin.
> 
> Note that when running tests going forward, we want to use the `--release` flag since the `SetupCache` `Arc<Mutex<..>>` is slow otherwise.

Please put these instructions in the readme and resolve the merge conflicts

src/basic_block/cqlin.rs
```diff
@@ -114,101 +114,125 @@ impl BasicBlock for CQLinBasicBlock {
...
-
-    let mut setup = R;
-    setup.append(&mut Q);
-    setup.append(&mut S);
-    setup.append(&mut P_R);
-    setup.append(&mut L_V_i_x_n);
-    setup.append(&mut L_V_i_x);
-    setup.append(&mut L_H_i_x);
-    (setup, vec![M_x.into()], Vec::new())
+    let M_x = if cache.get(&format!("cqlin_R_{:p}_{}", self, n)).is_none() {
```
ddkang, line 133: This seems pretty messy. Is there some way to combine this into a single block?

PR #163, opened 2025-04-21
Fixes for vae-decoder and unet

liliatangxy:
- Enable ReshapeBB to work when converting to view doesn't work
- Fix pow of 2 issues in old Conv, Mul, InstanceNorm
- Optimize out copy constraint when reshape is possible in old Conv, InstanceNorm, and MaxPool
- Fix end indexing in Slice


ddkang (commented):
Why don't we have tests for these if there were issues?

liliatangxy:
We never implemented layer testing. But I save all the ONNX layers I have been using for testing.

src/layer/gather.rs
```diff
@@ -34,11 +34,10 @@ impl Layer for GatherLayer {
     let mut graph = Graph::new();
     let mut indices_output = -2;
     // Avoid unwrapping a None value
-    if input_shapes[1].len() == 0 {
+    if input_shapes[1].len() == 0 || constants.len() > 1 && constants[1].is_some() {
```
Vimalakirti, line 37: Could you add comments for this case?

src/layer/lstm.rs
```diff
@@ -151,13 +151,6 @@ impl Layer for LSTMLayer {
       let _ = graph.addNode(change_SF_check, vec![(matmul_output, 0), (sublayer6, 0)]);
 
       // sublayer 7: MatMul for H_t and R_T
-      let matmul = graph.addBB(Box::new(RepeaterBasicBlock {
```
Vimalakirti: If we merge this into master, it might cause problem because m and n fields are eliminated.

liliatangxy: This was a mistake, I added it back

src/layer/gather.rs
liliatangxy: Updated the comment

Vimalakirti (approved):
If every layer has been tested, this PR looks good for me.

ddkang:
Why don't we have simple ONNX layer testing? Please open an issue for that...

liliatangxy:
Added #166 
All affected layers have been tested

Vimalakirti (approved):

ddkang (approved):

PR #164, opened 2025-04-21
Concat input shape

liliatangxy:
Add concat input shape field so that concatenated inputs are not separated by padding


src/tests.rs
```diff
@@ -217,8 +217,28 @@ fn testBasicBlocks() {
     &empty,
     &vec![&data_to_split],
   );
-  let data_to_concat = ArrayD::from_shape_fn(IxDyn(&[1, 2]), |_| Fr::rand(&mut rng));
-  testBasicBlock(ConcatBasicBlock { axis: 0 }, srs, &empty, &vec![&data_to_concat, &data_to_concat.clone()]);
+  let data_to_concat = ArrayD::from_shape_fn(IxDyn(&[1, 4]), |_| Fr::rand(&mut rng));
```
ddkang, line 220: Are these tests sufficient?

liliatangxy: Yes, we cover having multiple inputs and having inputs that aren't powers of 2 along the concatenated axis.

src/layer/concat.rs
```diff
@@ -88,20 +88,14 @@ impl Layer for ConcatLayer {
       // If not concatenating along the last axis, directly concatenate
       let mut constOfShape_shape = input_shapes[0].clone();
       constOfShape_shape[axis] = 1;
```
Vimalakirti: why do we still need this?

Vimalakirti:
I'm thinking of a special case that the current Concat may go wrong when its output goes into the shape field of the next layer. Below is an example:
In ONNX, we can use Concat to concatenate three inputs [1], [4], [4], which outputs [1,4,4]. Then, the [1,4,4] is fed into the `shape` field of the next layer (such as Expand). However, this goes wrong in the current implementation because the Concat layer will output [1, 4, 4, 0].

liliatangxy:
In this case, wouldn't Concat output the correct output shape of the `shape` field [3] so that Expand knows it is [1,4,4]?

In many other cases, inputs that indicate shapes may also be padded, so it may be helpful to double-check whether this is handled correctly at the layer level.

Vimalakirti:
> In this case, wouldn't Concat output the correct output shape of the `shape` field [3] so that Expand knows it is [1,4,4]?
> 
> In many other cases, inputs that indicate shapes may also be padded, so it may be helpful to double-check whether this is handled correctly at the layer level.

Good point, do all the related layers, which process `shape`, slice the shape first to make sure it doesn't go wrong?
I think it's good to check before we merge this PR. Except for this, the PR lgtm.

ddkang:
Why can't we have a bunch of small ONNX files to test edge cases like this?

liliatangxy:
There are two layers that take shapes as inputs. They perform 0 filtering to remove the 0 padding before doing operations in the specified locations in the code.

Expand: https://github.com/uiuc-kang-lab/zk-torch/blob/master/src/layer/expand.rs#L30

reshape: https://github.com/uiuc-kang-lab/zk-torch/blob/master/src/layer/reshape.rs#L41

I also tested on Expand and Reshape layers with 3-dimensional shapes and the output shapes are correct.

src/layer/lstm.rs
```diff
@@ -319,7 +319,11 @@ impl Layer for LSTMLayer {
     }
 
     // sublayer 22: Concat H_list
-    let concat = graph.addBB(Box::new(ConcatBasicBlock { axis: 0 }));
+    let shape = vec![1, num_directions, batch_size, hidden_size].iter().map(|&x| util::next_pow(x as u32) as usize).collect();
+    let concat = graph.addBB(Box::new(ConcatBasicBlock {
```
Vimalakirti, line 313: With the new ConcatBB, may we delete L311~318?

liliatangxy: Yes. I removed them

Vimalakirti (approved):

ddkang (approved):

PR #165, opened 2025-04-21
Add generalized Mira for MatMul and Permute

Vimalakirti:
As title stated.

After discussing with @liliatangxy today, we decide to merge the PRs in the following order to prevent the master branch from being broken. 
1. Add MatMul and Permute into `mira_para` (which is to merge this PR into PR #162)
2. Add CQ and CQ2 into `mira_para`
3. Add Repeater into `mira_para`
4. Merge `mira_para` into `master`

PR #162, opened 2025-04-17
Support generalized mira

Vimalakirti:
This PR updated the API interface of our accumulation scheme. 
Specifically, we do the following updates:
1. remove `acc_init`, which is simply a special case of `acc_prove`
2. add a new field (i.e., `Vec<PairingOutput<Bn<ark_bn254::Config>>>`) in `acc_proof`, which is the error term we used to reduce memory usage rather than store tremendous G1 and G2 elements.
3. support generalized Mira in the zk-torch paper, which is realized through the `AccProofLayout` defined in `util/fold.rs`
4. include the implementation of AccProofLayout for CQLin, Sum, and Mul. In the next PRs, we will first add MatMul and Permute; then add CQ and CQ2; and finally include RepeaterBB (and in Repeater, we have parallel Merkle accumulation)

PR #167, opened 2025-04-29
Add generalized Mira for CQ and CQ2

Vimalakirti:
As described in #165, this PR includes the parallelized Mira of CQ and CQ2.

### DIFF
--- a/src/basic_block/cq.rs
+++ b/src/basic_block/cq.rs
@@ -309,7 +309,7 @@ impl CQLayoutHelper {
     let err_1 = (
       vec![
         acc_1_g1.A_Q_x * acc_2.mu + acc_2_g1.A_Q_x * acc_1.mu,
-        acc_1_g1.A_x * acc_2_fr.Beta + acc_2_g1.A_x * acc_1_fr.Beta - acc_1_g1.M_x * acc_1.mu - acc_2_g1.M_x * acc_2.mu,
+        acc_1_g1.A_x * acc_2_fr.Beta + acc_2_g1.A_x * acc_1_fr.Beta - acc_1_g1.M_x * acc_2.mu - acc_2_g1.M_x * acc_1.mu,
         acc_1_g1.Part_C1.unwrap() * acc_2.mu
           + acc_2_g1.Part_C1.unwrap() * acc_1.mu
           + acc_1_g1.A_x_1.unwrap() * acc_2_fr.Model_r.unwrap()
--- a/src/basic_block/cq.rs
+++ b/src/basic_block/cq.rs
@@ -900,9 +900,9 @@ impl BasicBlock for CQBasicBlock {
 
     let beta = Fr::rand(rng);
     let mut result = beta == proof.2[0];
-    result &= model == proof.0[proof.0.len() - 2];
-    result &= input == proof.0[proof.0.len() - 1];
-    //assert!(result, "acc_proof for cq is not valid");
+    result &= model == proof.0[CQG1Terms::idx(CQG1Terms::Model_g1_blinded)];
+    result &= input == proof.0[CQG1Terms::idx(CQG1Terms::Input_g1_blinded)];
+    assert!(result, "acc_proof for cq is not valid");
     vec![]
   }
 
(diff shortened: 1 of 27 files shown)


src/basic_block/cq.rs
```diff
@@ -14,8 +23,622 @@ use rand::{rngs::StdRng, SeedableRng};
...
+    }
+  }
+
+  pub fn mira_prove(
+    srs: &SRS,
+    acc_1: AccHolder<G1Projective, G2Projective>,
+    acc_2: AccHolder<G1Projective, G2Projective>,
+    rng: &mut StdRng,
+  ) -> AccHolder<G1Projective, G2Projective> {
+    let m_x = acc_2.acc_g1[CQG1Terms::idx(CQG1Terms::M_x)];
```
ddkang: Why do we need to define these explicitly? Seems pretty bad...

Vimalakirti: If we don't define it explicitly, we will need to use indices to access these elements, which hurts the readability.

ddkang: why can't the macro make a struct and then you can do something like:
x = ...
x.variable_1
x.variable_2

src/basic_block/cq.rs
```diff
@@ -239,4 +883,93 @@ impl BasicBlock for CQBasicBlock {
...
+    _cache: ProveVerifyCache,
+  ) -> Vec<PairingCheck> {
+    let model = model.first().unwrap().g1;
+    let input = inputs[0].first().unwrap().g1;
+
+    let beta = Fr::rand(rng);
+    let mut result = beta == proof.2[0];
+    result &= model == proof.0[proof.0.len() - 2];
+    result &= input == proof.0[proof.0.len() - 1];
+    //assert!(result, "acc_proof for cq is not valid");
```
liliatangxy: I think we want to keep this check.

src/basic_block/cq2.rs
```diff
@@ -287,4 +384,95 @@ impl BasicBlock for CQ2BasicBlock {
...
+    let alpha = Fr::rand(rng);
+    let inputs = vec![inputs[0].first().unwrap(), inputs[1].first().unwrap()];
+    let agg_input: G1Affine = (inputs[0].g1 + (inputs[1].g1 * alpha)).into();
+    let agg_model: G1Affine = (model[0].g1 + (model[1].g1 * alpha)).into();
+
+    let beta = Fr::rand(rng);
+    let mut result = beta == proof.2[0];
+    result &= agg_model == proof.0[proof.0.len() - 2];
+    result &= agg_input == proof.0[proof.0.len() - 1];
+    //assert!(result, "acc_proof for cq2 is not valid");
```
liliatangxy: same with this check

liliatangxy (commented):
I think it looks ok minus the current comments

src/basic_block/matmul.rs
```diff
@@ -16,360 +18,272 @@ use ndarray::{arr1, arr2, ArrayD, Ix1, Ix2, IxDyn};
...
-  let acc_holder = matmul_acc_to_acc_holder(acc, is_prover);
-  holder_to_acc_proof(acc_holder)
+  fn mira_prove(
+    &self,
+    srs: &SRS,
+    acc_1: AccHolder<G1Projective, G2Projective>,
+    acc_2: AccHolder<G1Projective, G2Projective>,
+    rng: &mut StdRng,
+  ) -> AccHolder<G1Projective, G2Projective> {
+    let acc_flat_A = acc_1.acc_g1[MatMulG1Terms::idx(MatMulG1Terms::Flat_A)];
```
ddkang: why does this need the idx? Why can't you set up the macro so that you can refer to them directly by name?

src/basic_block/cq.rs
Vimalakirti: We can, I updated the macro in `fold.rs` for both PR #162 and this PR.

src/basic_block/matmul.rs
Vimalakirti: We can, I have updated the macro.

Vimalakirti:
@ddkang @liliatangxy Thanks for the valuable feedback. Please review this PR again when it is convenient for you.

ddkang (approved):

PR #168, opened 2025-05-02
Add generalized Mira for RepeaterBB

Vimalakirti:
As stated in #165 , this is the last PR in the generalized Mira, which supports generalized Mira in our RepeaterBB.

### DIFF
--- a/src/basic_block/repeater.rs
+++ b/src/basic_block/repeater.rs
@@ -1,11 +1,87 @@
-use super::{BasicBlock, Data, DataEnc, PairingCheck, ProveVerifyCache, SRS};
-use crate::{ndarr_azip, util};
+use super::{AccProofAff, AccProofAffRef, AccProofProj, AccProofProjRef, BasicBlock, Data, DataEnc, PairingCheck, ProveVerifyCache, SRS};
+use crate::basic_block::*;
+use crate::{
+  ndarr_azip,
+  util::{self, acc_proof_to_holder, holder_to_acc_proof, AccHolder, AccProofLayout},
+};
 use ark_bn254::{Fr, G1Affine, G1Projective, G2Affine, G2Projective};
+use ark_ec::bn::Bn;
+use ark_ec::pairing::{Pairing, PairingOutput};
 use ark_poly::univariate::DensePolynomial;
 use itertools::multiunzip;
 use ndarray::{arr1, azip, par_azip, s, ArrayD, Axis, Dimension, IxDyn, SliceInfo, SliceInfoElem};
 use rand::rngs::StdRng;
 use rayon::prelude::*;
+use std::sync::{Arc, Mutex};
+
+fn build_proven_level_sizes(mut n: usize) -> Vec<usize> {
+  let mut sizes = vec![n];
+  while n > 1 {
+    let pairs = n / 2; // only paired nodes produce new results
+    sizes.push(pairs);
+    n = (n + 1) / 2; // move to next level
+  }
+  sizes
+}
+
+macro_rules! downcast_to_layout {
+  ($bb:expr, $( $ty:ty ),+ ) => {
+    {
+      let bb_ref: &dyn AccProofLayout =
+        $(
+          if $bb.is::<$ty>() {
+            $bb.downcast_ref::<$ty>().unwrap() as &dyn AccProofLayout
+          } else
+        )+
+      {
+        &BasicBlockForTest {} as &dyn AccProofLayout
+      };
+      bb_ref
+    }
+  };
+}
+
+fn get_local_acc_proof_indices(
+  bb: &dyn BasicBlock,
+  acc_g1_len: usize,
+  acc_fr_len: usize,
+  is_prover: bool,
+) -> (Vec<usize>, Vec<usize>, Vec<usize>, Vec<usize>) {
+  if acc_fr_len == 0 {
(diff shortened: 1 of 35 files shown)


src/util/fold.rs
```diff
@@ -183,7 +187,7 @@ pub trait AccProofLayout: BasicBlock {
   }
 }
 
-pub fn acc_proof_to_holder<P: Copy, Q: Copy, L: AccProofLayout>(
+pub fn acc_proof_to_holder<P: Copy, Q: Copy, L: AccProofLayout + ?Sized>(
```
Vimalakirti, line 194: The syntax `L: ?Sized` means "`L` could be a type that implements `Sized` or a type that doesn’t implement `Sized`". This is required because we need this property to implement `get_acc_proof_bases(.)` in `repeater.rs`.

src/basic_block/repeater.rs
```diff
@@ -1,11 +1,72 @@
...
 use ark_poly::univariate::DensePolynomial;
 use itertools::multiunzip;
 use ndarray::{arr1, azip, par_azip, s, ArrayD, Axis, Dimension, IxDyn, SliceInfo, SliceInfoElem};
 use rand::rngs::StdRng;
 use rayon::prelude::*;
+use std::sync::{Arc, Mutex};
+
+// build the proven level sizes given the number of nodes
+// this is used to build the acc proof indices for the repeater verifier
+fn build_proven_level_sizes(mut n: usize) -> Vec<usize> {
```
ddkang: is this the best name we can think of?

src/basic_block/repeater.rs
```diff
@@ -1,11 +1,72 @@
...
+  while n > 1 {
+    let pairs = n / 2; // only paired nodes produce new results
+    sizes.push(pairs);
+    n = (n + 1) / 2; // move to next level
+  }
+  sizes
+}
+
+// downcast the basic block to the correct layout
+macro_rules! downcast_to_layout {
```
ddkang, line 43: what does this mean?

src/util/fold.rs
ddkang: why don't you leave this as a comment

src/basic_block/repeater.rs
```diff
@@ -203,4 +264,337 @@ impl BasicBlock for RepeaterBasicBlock {
...
+          ),
+          false,
+        )
+      })
+      .collect();
+    all_levels.append(&mut combined);
+
+    // Step 2: Merkle reduction level by level to perform the verification
+    let mut level_start = 0;
+    let lonely_child = Arc::new(Mutex::new(None));
```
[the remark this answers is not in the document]
Vimalakirti: no, we need the Mutex for thread safety.

src/basic_block/repeater.rs
Vimalakirti: added comments

src/util/fold.rs
Vimalakirti: added the comment

src/basic_block/repeater.rs
Vimalakirti: updated the name

src/basic_block/repeater.rs
```diff
@@ -1,11 +1,72 @@
...
+      let bb_ref: &dyn AccProofLayout =
+        $(
+          if $bb.is::<$ty>() {
+            $bb.downcast_ref::<$ty>().unwrap() as &dyn AccProofLayout
+          } else
+        )+
+      {
+        &BasicBlockForTest {} as &dyn AccProofLayout
+      };
+      bb_ref
```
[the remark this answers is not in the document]
Vimalakirti: Renamed the struct and added comments

src/basic_block/repeater.rs
ddkang: I can infer that from the function name... why do you need this...

src/basic_block/repeater.rs
ddkang: again, I don't know what this actually means beyond the function name...

Vimalakirti: This function can calculate the Merkle tree level sizes given the number of nodes.
We use this helper function for the following purpose:
The repeater prover uses Merkle reduction to prove the accumulation, where the root is the final accumulator instance. In the Merkle tree, all the intermediate nodes are the accumulator instances to be verified, which are stored in a vector. The repeater verifier needs to know how many nodes to process at each level to transform the vector into a Merkle tree for verification.

I added these as the comments above the function.

src/basic_block/repeater.rs
Vimalakirti: We use this macro for the following purpose:
The basic block in a repeater is a `Box<dyn BasicBlock>` (dynamic dispatch) and we need to transform it to the `AccProofLayout` defined in the basic block to access the acc proof methods including `mira_prove`, `prover_proof_to_acc`, etc. (they are not defined in `dyn BasicBlock`)

To access the methods defined in AccProofLayout of XXXBasicBlock, this macro can:
1. check if the basic block is a XXXBasicBlock
2. downcast the basic block to the XXXBasicBlock
3a. return the AccProofLayout of the XXXBasicBlock if we support acc proof for the XXXBasicBlock
3b. return the default AccProofLayout as a placeholder for other basic blocks that do not have an acc proof layout|

I added these as the comments above the macro.

ddkang (approved):

PR #169, opened 2025-05-04
Add customized Conv, Maxpool, Resize, and InstanceNorm

Vimalakirti:
As the title stated, this PR introduces the customized layer implementations that prevent us from copy constraints in CNNs.
I will mark this PR ready for review after I add some comments for readability and the python scripts (our compiler).


src/basic_block/new_conv.rs
```diff
@@ -0,0 +1,606 @@
...
+    Ok(vec![r])
+  }
+
+  fn encodeOutputs(&self, _srs: &SRS, _model: &ArrayD<Data>, inputs: &Vec<&ArrayD<Data>>, outputs: &Vec<&ArrayD<Fr>>) -> Vec<ArrayD<Data>> {
+    let C_o = self.out_channels;
+    let D = self.input_shape[0];
+    let H = self.input_shape[1];
+    let W = self.input_shape[2];
+
+    let k_d = self.kernel_shape[0];
```
ddkang: It this the best way to do this?

src/basic_block/new_conv.rs
```diff
@@ -0,0 +1,606 @@
...
+use rayon::prelude::*;
+
+/*
+  All basic blocks in this file are used to perform the customized addition of our
+  special convolutional layers (where we put the kernel dimension in the last dimension).
+  For instance, in the case of a 2D convolution, we can use CQLinBasicBlock to update the input tensor
+  from shape [1, H_in * W_in, C_in] to shape [1, H_in * W_in, C_out].
+  Then, we can use Conv2DAddBasicBlock here to perform the convolutional addition to map the tensor
+  from [1, H_in * W_in, C_out] to [1, H_out * W_out, C_out].
+*/
```
liliatangxy, line 8: Can we mention which cases are covered in the comment?

src/layer/new_conv.rs
```diff
@@ -0,0 +1,694 @@
...
+
+    let output_shape = vec![1, H_o * W_o, weight_shape[0]];
+    graph.outputs.push((output, 0));
+    (graph, vec![output_shape], vec![input_types[0]])
+  }
+}
+
+// This basic block is only used in RetinaNet where we need to concatenate the outputs of multiple conv heads along axis 1.
+#[derive(Debug)]
+pub struct MultiHeadConv2dAggBasicBlock {
```
liliatangxy, line 173: should this be in basic_block/new_conv?

src/basic_block/new_conv.rs
Vimalakirti: May I ask what you mean? I think these basicblocks are general. Do you mean we should mention what CNNs in MLPerf are proved by this?

src/layer/new_conv.rs
Vimalakirti: Thanks for the suggestion! I think it might be better to put it here since it is only used in the MultiHeadConv2dLayer, which is a special case in RetinaNet. A similar case appears in layer/gather.rs where GatherBasicBlock is put there for the same reason. But I may understand this inappropriately.

src/basic_block/new_conv.rs
Vimalakirti: updated the code with a shapeHelper

src/basic_block/new_conv.rs
liliatangxy: I saw that there are several assertion checks in the basic block implementation, e.g., starting on line 49. I think it's good to summarize those at the block comment

src/basic_block/new_conv.rs
liliatangxy: I think it might be better to have them as separate fields unless there is a time we want to loop over them.

Vimalakirti: Thanks for the comments, I updated it.

src/basic_block/new_conv.rs
Vimalakirti: I see, thanks for the advice. I added comments for the Conv3DTransposeBasicBlock, please check.

src/layer/resize.rs
```diff
@@ -90,3 +101,162 @@ impl Layer for ResizeLayer {
...
+    &self,
+    _srs: &SRS,
+    _model: &ArrayD<DataEnc>,
+    inputs: &Vec<&ArrayD<DataEnc>>,
+    outputs: &Vec<&ArrayD<DataEnc>>,
+    _proof: (&Vec<G1Affine>, &Vec<G2Affine>, &Vec<Fr>),
+    _rng: &mut StdRng,
+    _cache: ProveVerifyCache,
+  ) -> Vec<PairingCheck> {
+    //let inputs_g1 = inputs.iter().fold(G1Projective::zero(), |acc, x| acc + x.first().unwrap().g1);
```
liliatangxy: clean up comments

liliatangxy (dismissed):
lgtm

liliatangxy (approved):

Issue #136, opened 2024-10-21
Matmul condition on srs size and cqlin setup assertions


liliatangxy:
- Condition MatMul layer to use MatMulBB when srs isn't big enough to use CQLinBB
- Add warning when MatMul is using MatMulBB instead of CQLinBB when there is a fixed matrix

PR #171, opened 2025-08-06
Adding Matmul condition on srs size and cqlin setup assertions (Issue #136)

baptistapedro:
Added the constraints for issue #136 and tests

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,6 +28,7 @@ pub static CONFIG_FILE: Lazy<String> = Lazy::new(|| {
         args[1].clone()
     }
 });
+
 // Define a static CONFIG that holds the loaded configuration
 pub static CONFIG: Lazy<util::Config> = Lazy::new(|| {
   let mut file = File::open(&*CONFIG_FILE).expect("Could not open config");
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,13 +17,17 @@ use std::io::Read;
 use std::path::Path;
 
 pub static CONFIG_FILE: Lazy<String> = Lazy::new(|| {
-  let args: Vec<String> = env::args().collect();
-  if args.len() != 2 {
-    panic!("Usage: cargo run -- <config file>");
-  }
-  args[1].clone()
+    if let Ok(path) = std::env::var("ZK_TORCH_CONFIG") {
+        // For testing purposes
+        path
+    } else {
+        let args: Vec<String> = std::env::args().collect();
+        if args.len() != 2 {
+            panic!("Usage: cargo run -- <config file>");
+        }
+        args[1].clone()
+    }
 });
-
 // Define a static CONFIG that holds the loaded configuration
 pub static CONFIG: Lazy<util::Config> = Lazy::new(|| {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -17,13 +17,17 @@ use std::io::Read;
 use std::path::Path;
 
 pub static CONFIG_FILE: Lazy<String> = Lazy::new(|| {
-  let args: Vec<String> = env::args().collect();
-  if args.len() != 2 {
-    panic!("Usage: cargo run -- <config file>");
-  }
-  args[1].clone()
+    if let Ok(path) = std::env::var("ZK_TORCH_CONFIG") {
+        // For testing purposes
```
Vimalakirti, line 21: Is this feature intended to be added in this PR?

src/lib.rs
```diff
@@ -17,13 +17,17 @@ use std::io::Read;
...
+        path
+    } else {
+        let args: Vec<String> = std::env::args().collect();
+        if args.len() != 2 {
+            panic!("Usage: cargo run -- <config file>");
+        }
+        args[1].clone()
+    }
 });
-
```
Vimalakirti, line 26: Here we need an empty line.

Vimalakirti (commented):
Thank you for your contributions! I left some comments.

src/lib.rs
baptistapedro: Yes this is required, using that as an env var we can use directly the CONFIG file for testing considering that when running `cargo test` we cannot pass it as an `arg` as required in the original version.

src/lib.rs
baptistapedro: Done.

Vimalakirti (approved):

src/tests.rs
```diff
@@ -587,3 +617,163 @@ fn test_copy_constraint() {
...
+    // Setup common data
+    let input_values = load_sample_input();
+    let input_len = input_values.len();
+    let srs_log = crate::CONFIG.ptau.loaded_pow_len_log;
+    let srs_threshold = 1 << srs_log;
+    let a = input_len; // e.g., 4
+    let b = 64;
+    let matrix_size = a * b;
+
+    let input_vec_fr: Vec<Fr> = input_values.iter().map(|&x| f32_to_fr(x, 10)).collect();
```
liliatangxy, line 731: Can we simplify the test by generating a random Fr input and weights instead of loading floats from a file?

liliatangxy (commented):
Thanks for contributing to our codebase! CQLinBB is actually used in multiple types of layers (e.g., Conv, Matmul, Conv2d). I think instead of directly embedding the logic into the Matmul layer I think it would be better to make a helper function with this logic and call it for each usage of CQLinBB. This would also make testing cleaner, as we would be able to test the function directly instead of copy-pasting the code in the test.

src/tests.rs
baptistapedro: Glad to contribute to the project. I will change this logic into a helper function to apply to the other layers as soon as I get some free time.
Regarding simplifying the test, I thought that loading from a file would be more realistic feeding directly the files but shouldn't be a concern if changed that way.
Will do.

